PR #14, opened 2021-06-23
PCRM-3852 - Process and split supplementary data from DPS

abbas-khan8:
- TF scripts to create buckets: LR-20 and LR-22, and lambda: LR-21
- Functionality to read a csv file from the input bucket, splitting on GP code and writing individual files to output bucket


lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,211 @@
...
+
+        return data_lines
+
+    except (ClientError, UnicodeDecodeError) as err:
+        msg = f"LR-21 failed to read supplementary data from: {upload_key}"
+        error_response = log_dynamodb_error(
+            "99999999-2121-2121-2121-999999999999", "HANDLED_ERROR", msg
+        )
+
+        raise type(err)(error_response) from err
```
jamesnunnnhs, line 100: I forgot to say that we should actually not use type(err) because some exceptions require extra args - can you raise a custom exception or Exception instead? 

`Exception(error_response) from err`

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,207 @@
...
+    except ValueError as err:
+        msg = f"LR-21 failed to process file contents for: {upload_key}"
+        error_response = log_dynamodb_error(
+            "99999999-2121-2121-2121-999999999999", "HANDLED_ERROR", msg
+        )
+
+        raise InvalidDSAFile(error_response) from err
+
+
+def write_files(per_registered_gp: defaultdict):
```
jamesnunnnhs, line 139: Since you're not using any features of `defaultdict`, don't require it and instead just accept a `Dict`.

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,207 @@
...
+        stream = io.StringIO()
+
+        writer = csv.DictWriter(stream, fieldnames=headers, lineterminator="\n")
+        writer.writeheader()
+        writer.writerows(per_registered_gp[gp])
+
+        csv_results_string = stream.getvalue().strip()
+
+        try:
+            client.put_object(Body=csv_results_string, Bucket=OUTPUT_BUCKET, Key=f"{gp}.csv")
```
jamesnunnnhs, line 160: The file will contain 60m records and thousands of practices - have you tested how long it take to process that many?

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,207 @@
...
+        pages = paginator.paginate(Bucket=OUTPUT_BUCKET)
+
+        incoming_files = set(per_registered_gp.keys())
+        existing_files = set()
+
+        for page in pages:
+            for obj in page["Contents"]:
+                existing_files.add(str(obj["Key"]).replace(".csv", ""))
+
+        outdated_files = [existing for existing in existing_files if existing not in incoming_files]
```
jamesnunnnhs, line 193: Since you're using `set`s you can do this instead `outdated_files = existing_files - incoming_files`

However didn't Ben say to delete everything in the bucket regardless of incoming practices?

abbas-khan8: I don't remember exactly, I'll message him to confirm I think the point was to treat new data as the the only data to use and ignore/delete everything else.

lambdas/LR_21_split_dps_extract/split_dps_extract.py
abbas-khan8: Not with 60m yet, I've created a mockaroo and modified your mock_data script to work with it. It's been running for a while so once that finishes calling the API I can test it

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,207 @@
...
+        "s3",
+        region_name=REGION,
+        aws_access_key_id=ACCESS_KEY,
+        aws_secret_access_key=SECRET_KEY,
+        aws_session_token=SESSION_TOKEN,
+    )
+
+
+def lambda_handler(event, context):
+    """LR-21 lambda invocation handler"""
```
josh-scargill-nhsd, line 36: irrelevant comment

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,207 @@
...
+    except Exception as err:
+        msg = f"Unhandled error when processing supplementary data file in LR-21"
+        error_response = log_dynamodb_error(
+            "99999999-2121-2121-2121-999999999999", "UNHANDLED_ERROR", msg
+        )
+
+        raise Exception(error_response) from err
+
+
+def split_dps_extract_handler(upload_key: str) -> success:
```
josh-scargill-nhsd, line 52: This doesn't need to have the word handler does it? 🤔

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,207 @@
...
+    pass
+
+
+def create_s3_client() -> BaseClient:
+    return boto3.client(
+        "s3",
+        region_name=REGION,
+        aws_access_key_id=ACCESS_KEY,
+        aws_secret_access_key=SECRET_KEY,
+        aws_session_token=SESSION_TOKEN,
```
josh-scargill-nhsd, line 31: Are these all needed? i've been doing this without these passed in and it works fine 🤷

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,207 @@
...
+    Args:
+        upload_key (str): A path to the s3 file key object
+        per_registered_gp (defaultdict): A dict of registered GP's
+    """
+
+    client = create_s3_client()
+
+    try:
+        paginator = client.get_paginator("list_objects_v2")
+        pages = paginator.paginate(Bucket=OUTPUT_BUCKET)
```
josh-scargill-nhsd, line 184: Just to clarify, we can't just, on request, pass in a condition too?

pytest.ini
```diff
@@ -15,6 +15,11 @@ env =
     INFLIGHT_TABLE=In_Flight_Table
     AWS_S3_REGISTRATION_EXTRACT_BUCKET=mock-pds-data
     AWS_PATIENT_RECORD_SQS=mock-test-queue
+    LR_20_SUPPLEMENTARY_INPUT_BUCKET=mock-dps-input-bucket
+    LR_22_SUPPLEMENTARY_OUTPUT_BUCKET=mock-dps-output-bucket
+    AWS_ACCESS_KEY_ID=TESTID
+    AWS_SECRET_ACCESS_KEY=TESTKEY
+    AWS_SESSION_TOKEN=TESTTOKEN
```
josh-scargill-nhsd, line 22: won't need these if you remove the above bit

terraform/modules/lambdas/LR-21/main.tf
```diff
@@ -0,0 +1,16 @@
...
+  principal     = "s3.amazonaws.com"
+  source_arn    = "arn:aws:s3:::${var.supplementary-input-bucket}"
+}
+
+resource "aws_s3_bucket_notification" "bucket_notification" {
+  bucket = var.supplementary-input-bucket
+
+  lambda_function {
+    lambda_function_arn = aws_lambda_function.LR-21-Lambda.arn
+    events              = [ "s3:ObjectCreated:*"]
```
josh-scargill-nhsd, line 14: silly question, but it only triggers on create right? not removal etc
(As in you've tested this lol)

abbas-khan8: I've not tested it but judging by the docs [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html), there's a separate trigger for deleting/removal

lambdas/LR_21_split_dps_extract/split_dps_extract.py
abbas-khan8: The logic behind that was that It's a function which handles all the other function calls, so I called it a handler. You have another name in mind?

lambdas/LR_21_split_dps_extract/split_dps_extract.py
abbas-khan8: Not sure I understand your q

lambdas/LR_21_split_dps_extract/split_dps_extract.py
abbas-khan8: I think it's a force of habit, if yours works without then I'm sure mine will too

pytest.ini
abbas-khan8: These are required for the parser too, not just the lambda

lambdas/LR_21_split_dps_extract/split_dps_extract.py
josh-scargill-nhsd: get all bucket files where last modified date < datetime.now()-4 hours

terraform/modules/lambdas/LR-21/main.tf
josh-scargill-nhsd: just do a quick test, will take 2 mins to establish

jamesnunnnhs (approved):

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,198 @@
...
+
+        for page in pages:
+            for obj in page["Contents"]:
+                existing_files.add(str(obj["Key"]).replace(".csv", ""))
+
+        outdated_files = existing_files - incoming_files
+
+        if outdated_files:
+            for gp_file in outdated_files:
+                client.delete_object(Bucket=OUTPUT_BUCKET, Key=f"{gp_file}.csv")
```
aparimoo-nhs, line 188: `outdated files` needs to be logged before it is being deleted.

lambdas/LR_21_split_dps_extract/split_dps_extract.py
abbas-khan8: I found this [here](https://github.com/boto/boto3/issues/949) but I don't think it will work, since it uses JMESpath and  compares strings instead of datetime objects

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,198 @@
...
+
+        client.delete_object(Bucket=INPUT_BUCKET, Key=upload_key)
```
aparimoo-nhs, line 190: Why are you deleting this?  If all the files are deleted from the folder then you don't  need to do this ?

abbas-khan8: This is the original supplementary file which is being deleted from the inbound folder, the other files are deleted in the output folder

lambdas/LR_02_validate_and_parse/validate_and_parse.py
```diff
@@ -159,23 +150,19 @@ def handle_validated_records(
...
-        job_id: str,
-        gp_ha_cipher: str,
-        upload_filename: str,
-        records: list,
-        num_of_records: int
+    job_id: str,
+    gp_ha_cipher: str,
+    upload_filename: str,
+    records: list,
+    num_of_records: int,
```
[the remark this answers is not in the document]
abbas-khan8: Black applies commas at the end of lists on formatting

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,198 @@
...
+    lines = read_file(upload_key)
+    LOG.info("Data lines extracted")
+
+    per_registered_gp = split_file(upload_key, lines)
+    LOG.info("Dictionary created for each registered GP")
+
+    write_files(per_registered_gp)
+    LOG.info("Output files for each registered GP was successful")
+
+    cleanup_files(upload_key, per_registered_gp)
```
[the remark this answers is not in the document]
abbas-khan8: It could be but it doesn't just cleanup the outdated files, but also the original supplementary file once consumed

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,198 @@
...
+    """Handler function for DPS file split process
+
+    Args:
+        upload_key (str): A path to the s3 file key object
+
+    Returns:
+        success: A dict result containing a status and message
+    """
+
+    lines = read_file(upload_key)
```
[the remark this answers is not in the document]
abbas-khan8: Sure, they are not valid records just yet they are just file lines at this point though

lambdas/LR_21_split_dps_extract/split_dps_extract.py
abbas-khan8: There will potentially be hundreds, maybe even thousands of GP practices being affected here. I don't know if logging this many out would be a good idea?

lambdas/LR_02_validate_and_parse/validate_and_parse.py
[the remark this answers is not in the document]
aparimoo-nhs: strange , didnt know that

lambdas/LR_21_split_dps_extract/split_dps_extract.py
aparimoo-nhs: question for operations i believe. They would want to know what files are deleted in case of any accidents

lambdas/LR_21_split_dps_extract/split_dps_extract.py
josh-scargill-nhsd: I forget where we are with this, I think we discussed in Slack afterwards, if this has been sorted let me know 👍

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,197 @@
...
+
+def lambda_handler(event, context):
+    """LR-21 lambda invocation handler
+
+    Args:
+        event (dict): Uploaded S3 object key
+
+    Returns:
+        success: A dict result containing a status and message
+    """
```
josh-scargill-nhsd, line 29: Not convinced this is needed, we know what a lambda handler does :D

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,197 @@
...
+        msg = f"Unhandled error when processing supplementary data file in LR-21"
+        error_response = log_dynamodb_error(
+            "99999999-2121-2121-2121-999999999999", UNHANDLED_ERROR, msg
+        )
+
+        raise Exception(error_response) from err
+
+
+def split_dps_extract(upload_key: str) -> success:
+    """Handler function for DPS file split process
```
josh-scargill-nhsd, line 46: You've renamed the function so that's half the file, let's drop handler from here too, what about something more like "Splits a single file into multiple smaller files by practice", or something along those lines?

lambdas/LR_21_split_dps_extract/split_dps_extract.py
```diff
@@ -0,0 +1,197 @@
...
+    Returns:
+        list: A list of strings containing file data
+    """
+
+    try:
+        client = boto3.client("s3")
+
+        file_obj = client.get_object(Bucket=INPUT_BUCKET, Key=upload_key)
+
+        file_data = file_obj["Body"].read().decode("utf-8").splitlines()
```
josh-scargill-nhsd, line 85: I could be crazy but i'm sure i've seen @jamesnunnnhs make a generic thing for this. It might be in a PR, if it is, and it's suitable here, have him update his PR since yours will beat it in :)

packages/utils/datetimezone.py
```diff
@@ -26,4 +26,4 @@ def localize_date(unlocalized_date: datetime = get_datetime_now(), specified_tim
             datetime: Localized datetime
     """
 
-    return timezone(specified_timezone).localize(unlocalized_date)
+    return unlocalized_date.replace(tzinfo=utc).astimezone(timezone(specified_timezone))
```
josh-scargill-nhsd, line 29: If there isn't a test for this, and you originally wrote it, make sure theres tests to ensure it's working as expected

pytest.ini
josh-scargill-nhsd: If they're required for the parser, why are you only just adding them now? 🤔

test/unittests/lambdas/LR_21_split_dps_extract/test_LR_21_split_dps_extract.py
```diff
@@ -0,0 +1,163 @@
...
+
+    updated_gp_files = []
+
+    for page in pages:
+        for obj in page["Contents"]:
+            updated_gp_files.append(obj["Key"])
+
+    assert EXISTING_GP_FILE in updated_gp_files
+
+def test_invalid_file_raises_value_error(upload_invalid_dps_data_to_s3, create_dynamodb_tables):
```
josh-scargill-nhsd, line 161: 2 lines above function

pytest.ini
abbas-khan8: Because we hadn't added moto tests until recently

lambdas/LR_21_split_dps_extract/split_dps_extract.py
abbas-khan8: I have contacted Mr Nunn, and he assures me that you're just crazy

josh-scargill-nhsd (approved):

PR #15, opened 2021-06-24
PCRM-3570 - Fixes trailing dash when suffix is blank string

jamesnunnnhs:
Fixes trailing dash when suffix is blank string for preprod, test and prod environments


terraform/environment/dev/main.tf
```diff
@@ -28,7 +28,7 @@ module "List-Recon" {
   pds_url = "pds_api_data.csv"
   patient_sqs = "Patient_Records.fifo"
   runtime = var.runtime
-  suffix = "${element(split("-", terraform.workspace), 1)}"
+  suffix = "-${element(split("-", terraform.workspace), 1)}"
```
aparimoo-nhs, line 31: As suggested  can we try this 
`suffix = terraform.workspace == "test" && terraform.workspace == "preprod" ? terraform.workspace:element(split("-", terraform.workspace), 1)`

aparimoo-nhs: May be as part of refactoring, this syntax may be wrong. @joeds13  your thoughts on this ?

aparimoo-nhs (approved):
Can you please update Readme as per new Makefile

joeds13:
Superseded by #16

terraform/environment/dev/main.tf
josh-scargill-nhsd: `terraform.workspace == "test" && terraform.workspace == "preprod"`
Should that be an `||`? it would never be both right?

josh-scargill-nhsd (approved):

PR #16, opened 2021-06-24
PCRM-3781 - Github Actions Private Runner

joeds13:
This will move the Github Actions Workflows to run against the NHSD AWS Accounts, CI builds will no longer run into Answer Digitals AWS Accounts

Using a Github Actions Private runner to deploy terraform from inside the AWS account

Sets the environment name to be of the format `PCRM-00000-abcd01234` where `PCRM-00000` is the ticket number from the branch and `abcd01234` is the first 8 digits of the a sha1 hash of the branch name, this provides a unique build env name for every branch but still makes them recognisable to humans.


.github/workflows/create-feature-branch-infra.yml
```diff
@@ -0,0 +1,53 @@
...
+      FEATURE_BRANCH: ${GITHUB_REF##*/}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Python Package
+        run: |
+          make python-package
+      - name: Terraform Init
+        run: |
+          make init
+          make workspace terraform_workspace=$(echo ${{ env.FEATURE_BRANCH }} | grep -Eo '^(\w+/)?(\w+[-_])?[0-9]+' | grep -Eo '(\w+[-])?[0-9]+' | tr "[:lower:]" "[:upper:]")-$(echo ${{ env.FEATURE_BRANCH }} | sha1sum | head -c 8)
```
jamesnunn, line 22: So given a branch like PCRM-1234-my-branch this produces `PCRM-1234-60a9dc5a` - did you mean to keep the `PCRM` portion?

makefile
```diff
@@ -1,30 +1,43 @@
...
-system_test: all
-	cd test/integrationtests && gauge run
-	cd ../..
+integrationtest-deps:
+	pip install -r test_requirements.txt
+	pip install -e .
 
-unittest: all
-	pytest
+integrationtests:
```
jamesnunn, line 39: Is it worth including integrationtest-deps as a dependency for this since it seems it's run every time integrationtests is run?

.github/workflows/create-feature-branch-infra.yml
joeds13: Personally I think it makes it easier for a person to recognise so left it in

makefile
```diff
@@ -1,30 +1,43 @@
...
+all: python-package init deploy
+test: integrationtests unittest
+
+python-package:
 	rm -r ./lambda_layer || true
 	mkdir -p ./lambda_layer/python/lib/python3.8/site-packages
 	docker run --rm -v $(PWD):/var/task -u $(shell id -u):$(shell id -g) -w="/var/task/" public.ecr.aws/sam/build-python3.8 /bin/python3 -m pip install -r requirements.txt -t ./lambda_layer/python/lib/python3.8/site-packages
 	cp -r ./packages/* ./lambda_layer/python/lib/python3.8/site-packages
-	terraform -chdir=terraform/environment/${deployment_environment} init
+
```
jamesnunn, line 12: For the purposes of testing, should include 
```
pip install -r test_requirements.txt
pip install .```

File: makefile (hunk 2 above)
Comment (joeds13) [reply makefile:39]: I left it off as a dependency it'll be a waste of time/resources in local dev if you're just wanting to run the tests repeatedly

File: makefile (hunk 3 above)
Comment (joeds13) [reply makefile:12]: Is that not covered off by the `integrationtest-deps`?

Comment (jamesnunn) [reply makefile:12]: So it is, Still thinking then

Comment (joeds13) [reply makefile:12]: Fixed by setting the PATH in the CI

**Comment by joeds13** [comment]
@premnhs confirmed to use a tag to stop the problematic running until they can look into it

**Review by jamesnunnnhs — APPROVED** [review]
Looks good

**Comment by jamesnunnnhs** [comment]
Worth nothing that the suffix added to feature branch stacks means that where any character length limits apply to aws service names, they're reduced by 20. So for example I believe now bucket names set in terraform are (63-20) 43 max chars long to allow for the suffix. Should be plenty but just sayin

File: .github/workflows/create-feature-branch-infra.yml [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/upload-artifact@v2
+        with:
+          name: terraform output
+          path: output.json
+          retention-days: 2
+      - name: Assume Role in Dev Account for Integration Tests
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: eu-west-2
+          role-to-assume: "arn:aws:iam::092420156801:role/LRTerraformDeploy"
```
josh-scargill-nhsd, line 36: do we want this stored in variables?

terraform/environment/dev/locals.tf
```diff
@@ -0,0 +1,14 @@
+locals {
+  name               = "list-reconciliation-${local.environment}"
+  environment        = lower(terraform.workspace)
+
+  tags = {
+    TagVersion         = "1"
```
josh-scargill-nhsd, line 6: Do we need version? do we ever intend to change it? I've always found version numbers can be redundant in systems like this

terraform/stacks/management/README.md
```diff
@@ -0,0 +1,3 @@
+# Management Stack
+
+Terraform State/Lock and Github Actions Private Runner
```
josh-scargill-nhsd, line 3: Is this meant to contain more?

aparimoo-nhs (commented):
LGTM

terraform/environment/dev/locals.tf
joeds13: Those tags are Spine's expected tag set, so I dropped them in as provided for consistency

terraform/stacks/management/README.md
joeds13: Yeah eventually, I originally just had it as a placeholder in the empty directory, as we fill out terraform resources in the stack we can add more info in

.github/workflows/create-feature-branch-infra.yml
joeds13: As this is the feature branch workflow we will only ever want to use this one role, if we had a single workflow file that worked for all environments we might want to move this to a variable so we can programatically access the desired value

josh-scargill-nhsd (approved):

PR #17, opened 2021-06-25
PCRM-3517 - Output demographics differences DSA work items

jamesnunnnhs:
Creates one or more work items in a bucket, one per patient for a job, where one or more demographics differences has been identified for a patient.


lambdas/LR_15_process_demographic_differences/process_demographic_differences.py
```diff
@@ -0,0 +1,212 @@
...
+
+    except Exception as err:
+        msg = f"Unhandled error getting PDS registrations. JobId: {job_id or '99999999-0909-0909-0909-999999999999'}"
+        error_response = log_dynamodb_error(job_id, UNHANDLED_ERROR, msg)
+
+        raise Exception(error_response) from err
+
+
+def create_dsa_payload(
+    patient_record: Dict, demo_diffs: List[Dict]
```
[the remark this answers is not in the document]
jamesnunnnhs: This makes it implementation agnostic - if we moved away from pynamo then we wouldn't have to change this.

lambdas/LR_15_process_demographic_differences/process_demographic_differences.py
```diff
@@ -0,0 +1,212 @@
...
+    for diff in demo_diffs:
+        patients[diff.PatientId].append(diff.attribute_values)
+
+    LOG.info(f"Processing {len(patients)} patients for JobId {job_id}")
+
+    job_pds_updated = 0
+    job_gp_updated = 0
+    job_human_validations = 0
+    job_potential_pds_updates = 0
+    job_potential_gp_updates = 0
```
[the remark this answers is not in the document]
jamesnunnnhs: you mean a single object containing multiple counters? It'd be the same amount of code or longer because we still need these specific counters, plus the setup code.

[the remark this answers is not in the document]
josh-scargill-nhsd: But these are the same properties as the JobStats class, so we already have the class shape here no?

lambdas/LR_15_process_demographic_differences/process_demographic_differences.py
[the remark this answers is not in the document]
josh-scargill-nhsd: But for now we're under the assumption pynamo is here to stay. And with that in mind we'd want to pass in the class. This also means if we renamed any properties it would be easier to catch

lambdas/LR_15_process_demographic_differences/process_demographic_differences.py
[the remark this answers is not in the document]
jamesnunnnhs: Yes but if you see further down in the code how I deal with getting/setting the job stats object, it's not exactly straightforward or tidy to update the model object in place. I think it is better to update these variables separately then just update the model in one go at the end.

lambdas/LR_15_process_demographic_differences/process_demographic_differences.py
```diff
@@ -0,0 +1,212 @@
...
+                JobStats.PotentialPdsUpdateRecords.set(job_potential_pds_updates),
+                JobStats.PotentialGpUpdateRecords.set(job_potential_gp_updates),
+            ]
+        )
+    LOG.info("Job stats updated")
+
+    job = get_job(job_id)
+    job.update(actions=[Jobs.StatusId.set(JobStatus.DEMOGRAPHICS_DIFFERENCES_PROCESSED.value)])
+
+    LOG.info("Job updated")
```
aparimoo-nhs, line 207: perhaps `job updated for {job_id}`

lambdas/LR_15_process_demographic_differences/process_demographic_differences.py
```diff
@@ -0,0 +1,212 @@
...
+        s3 = boto3.client("s3")
+        s3.put_object(Bucket=MESH_SEND_BUCKET, Key=key, Body=json.dumps(dsa_item))
+        out_files.append(f"s3://{MESH_SEND_BUCKET}/{key}")
+
+    LOG.info(f"Done")
+
+    try:
+        job_stat = JobStats.get(job_id)
+
+    except JobStats.DoesNotExist:
```
aparimoo-nhs, line 182: ` except Exception` ?

packages/utils/models.py
```diff
@@ -31,6 +31,16 @@ class Meta:
     JobId = UnicodeAttribute(hash_key=True)
 
 
+class DemographicsDifferencesJobIdIndex(GlobalSecondaryIndex):
+    class Meta:
+        index_name = "demographicsdifferences-job_id-index"
```
aparimoo-nhs, line 36: hardcoded ? , could it be env variable

terraform/modules/list-rec/buckets.tf
```diff
@@ -74,4 +74,14 @@ resource "aws_s3_bucket" "mock-pds-data" {
   tags={
     Name = "File storage for PDS mock api data"
   }
+}
+
+resource "aws_s3_bucket" "LR-23" {
+  bucket = "lr-22-mesh-send-bucket-${var.suffix}"
```
aparimoo-nhs, line 80: can it be shortened like `lr-22-output-bucket `

lambdas/LR_15_process_demographic_differences/process_demographic_differences.py
```diff
@@ -0,0 +1,212 @@
...
+
+    diffs = []
+
+    human_validations = 0
+    pds_updated = 0
+    gp_updated = 0
+    potential_pds_updates = 0
+    potential_gp_updates = 0
+
+    for demo_diff in demo_diffs:
```
aparimoo-nhs, line 59: consider renaming variable `demo_diff` to `demographic_diff` , it appears that code is referring to demo, but it could me

aparimoo-nhs (approved):

abbas-khan8 (approved):

jamesnunnnhs:
Just occurred to me, I'm waiting for Noor to update me on the filenames required for the outputs so I'll mark as WIP until I have the spec

lambdas/LR_08_demographic_comparison/demographic_comparison.py
```diff
@@ -10,8 +9,6 @@
 
 
 def lambda_handler(event, context):
-    from demographic_comparison import demographic_comparisons
-
     job_id = str(event["job_id"])
     patient_id = str(event["patient_id"])
```
abbas-khan8, line 13: Should these not exist inside the `try/catch`? If the event dict does not contain the correct information, this will crash right? I get this would break the `msg` but since it's a broad unhandled exception, wouldn't it be better to put "Unhandled error in when performing demographic comparison in LR-08" or something along those lines?

lambdas/LR_15_process_demo_diffs/process_demographic_differences.py
```diff
@@ -0,0 +1,218 @@
...
+from utils.statuses import JobStatus
+from services.jobs import get_job
+
+
+MANUAL_VALIDATION = "Manual Validation"
+MESH_SEND_BUCKET = os.getenv("MESH_SEND_BUCKET")
+
+
+def lambda_handler(event, context):
+    job_id = event["job_id"]
```
abbas-khan8, line 21: Should this not exist inside the try/catch? If the event dict does not contain the correct information, this will crash right? I get this would break the msg but since it's a broad unhandled exception, wouldn't it be better to put "Unhandled error in when performing demographic comparison in LR-08" or something along those lines? (Same for the other handlers too)

PR #18, opened 2021-06-28
PCRM-3893 - Adds PDS sensitive flag

jamesnunnnhs:


test/unittests/packages/utils/test_pds_api_service.py
```diff
@@ -0,0 +1,71 @@
...
+            CreateBucketConfiguration={"LocationConstraint": AWS_REGION},
+        )
+        s3.upload_file(os.path.join(DATA, "pds_api_data.csv"), bucket, "pds_api_data.csv")
+        yield
+
+
+def test_get_pds_fhir_record():
+
+    actual = get_pds_fhir_record(
+        "https://sandbox.api.service.nhs.uk/personal-demographics/FHIR/R4/Patient", "9000000009"
```
aparimoo-nhs, line 33: could this URL be as constant or fixture, i believe this will be used across tests

aparimoo-nhs (approved):

abbas-khan8 (approved):

test/unittests/packages/utils/test_pds_api_service.py
```diff
@@ -0,0 +1,71 @@
...
+        "address": ["1 Trevelyan Square", "Boar Lane", "City Centre", "Leeds", "West Yorkshire"],
+        "postcode": "LS1 6AE",
+        "gp_code": "Y12345",
+        "gp_registered_date": "2020-01-01",
+        "sensitive": "U",
+        "version": "2",
+    }
+
+    assert actual == expected
+
```
josh-scargill-nhsd, line 51: This test is fine to have, as we should be testing it's functionality, but our test shouldn't actually be relying on the response from the api. 2 reasons from me as for why:
1) Everytime local devs or the CI run the tests we'd be making this call to their actual sandbox environment
2) If at point the API changes or goes down, we won't be able to achieve 100% passing of tests. It'll fail our CI builds etc

jamesnunnnhs: Fair enough, we keep the test but what alternative do you suggest? Skip the test and only run it when explicitly requested?

josh-scargill-nhsd: My instinct is to fake the response for that call, not sure how viable that is with Python, can skip for now if that is something that we have no experience in doing.

aparimoo-nhs: If API changes or goes down tests fail , isn't what tests are for ? We should test the API contract

josh-scargill-nhsd: Would something like this work: https://bhch.github.io/posts/2017/09/python-testing-how-to-mock-requests-during-tests/ ?

josh-scargill-nhsd: We shouldn't be testing their API uptime no, that's not something we can control

jamesnunnnhs: But hold on - if the api changes then we _want_ to know about it sooner rather than later. Sandbox is for development anyway so it's expected to be used? It's not like we're hitting it multiple times a second.

jamesnunnnhs: API uptime is a different thing to API response content. I think we can handle any unlikely downtime that occurs _just_ at the time we're running tests

josh-scargill-nhsd: I'm not convinced the method we should be kept up to date with the APIs changes is our tests, but I also understand the prediciment.

I think we should fake the response, then raise this as an issue for the business side to tackle "If the API content changes, how are we supposed to keep up to date with it?"

josh-scargill-nhsd: There was a note someone just made to me, we could use integration tests to validate if the content has changed to what we're expecting, so we could look at that, just ensure the builds don't fail if it doesn't pass.
It's just this wouldn't live as a unit test

josh-scargill-nhsd: Could speak with @premnhs about setting up a `contract-tests` section that we'd run on a nightly basis or something?

josh-scargill-nhsd (approved):

josh-scargill-nhsd (approved):

PR #23, opened 2021-06-29
PCRM-3856 - Determine GP code from flat file

abbas-khan8:
- Refactor parser to accommodate new filename change, and return a dictionary for each valid file
- Write new GP code field to Jobs and Demographics table in LR-02


packages/gp_file_parser/file_name_parser.py
```diff
@@ -10,30 +9,25 @@ class InvalidFilename(Exception):
...
-        InvalidFilename: Raised when filename doesn't match expressions/
-                         Raised when invalid dates contained within filename
+        InvalidFilename: Raised when filename doesn't match correct format
     """
 
     filename = filename.upper()
 
-    valid_name = re.search(r"^GPR4([A-Z0-9]{3})1", filename)
+    valid_name = re.search(
+        r"^([A-HJ-NP-WY][LUE0-9][DER0-9][0-9][0-9][0-9])_GPR4([A-Z0-9]{3})1", filename
```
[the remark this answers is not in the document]
abbas-khan8: Used the docs [here](https://digital.nhs.uk/services/organisation-data-service/data-downloads/gp-and-gp-practice-related-data) for validation

packages/gp_file_parser/parser.py
```diff
@@ -469,22 +471,25 @@ def parse_gp_extract_file_s3(
         "s3",
         aws_access_key_id=os.environ.get("AWS_ACCESS_KEY_ID"),
         aws_secret_access_key=os.environ.get("AWS_SECRET_ACCESS_KEY"),
-        aws_session_token=os.environ.get("AWS_SESSION_TOKEN")
+        aws_session_token=os.environ.get("AWS_SESSION_TOKEN"),
```
josh-scargill-nhsd, line 474: Wild question, if you didn't have the keys declared, what happens? my impression is it'd still work.
Moreover, just thinking about it, this is going to fail on peoples machines if you use the credentials file to develop in without environment variables.

I think moving forward we should not declare these, and let boto3 figure it out (unless there is a reason we shouldn't be doing this that i'm not aware of)

packages/gp_file_parser/file_name_parser.py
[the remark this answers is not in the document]
jamesnunnnhs: Can you be more specific and maybe just reference it somewhere so it can be checked now or if there's any bugs later?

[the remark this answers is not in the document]
abbas-khan8: Spoken to Noor, the doc above is outdated and the new requirements have been added. Mention in the docstring

josh-scargill-nhsd:
Ben raised an issue about what returning different error messages based on name, can you pick that up as part of this?

abbas-khan8:
> Ben raised an issue about what returning different error messages based on name, can you pick that up as part of this?

@josh-scargill-nhsd This has been added as part of the refactoring

packages/gp_file_parser/file_name_parser.py
```diff
@@ -1,51 +1,93 @@
...
+        r"^([A-Z][0-9][0-9][0-9][0-9][0-9])", filename
+    )
+
+    if not valid_practice_code:
+        filename_errors.append("Filename does not contain a valid practice code")
 
-    if not valid_name:
-        raise InvalidFilename("Filename must have the correct format")
+    valid_download_name = re.search(
+        r"_GPR4([A-Z0-9]{3})1", filename
```
josh-scargill-nhsd, line 41: Could this be confusing? If I actually submitted `A11111-GPR4AAA1` it would say the filename does not contain a failed DOW name, when in fact i used the wrong separator

Maybe a simple check for if the filename doesn't contain `_` then it doesn't contain the correct separator

Also I wonder if we should, instead of returning multiple errors per file for the name, just return 1, based on severity.
```
Cannot detect separator
Invalid Practice Code
Invalid Filename
Invalid File extension
```

I'm not sure if how you're doing it at the moment you could be throwing red herrings, and if you can really avoid it

test/unittests/packages/gp_file_parser/test_filename_parser.py
```diff
@@ -33,42 +33,58 @@ def test_validate_filename_valid_lowercase_filename_returns_valid_date():
...
+    with pytest.raises(InvalidFilename):
+        validate_filename(filename)
+
+
+@freeze_time("2020-02-01")
+def test_validate_filename_invalid_gp_code_and_invalid_filename_raises_InvalidFilename():
+    filename = "AZ2023_GAR4BRF1.B1A"
+
+    with pytest.raises(InvalidFilename):
+        validate_filename(filename)
```
josh-scargill-nhsd, line 58: I'm expecting you to assert that the message contains what you're expecting, we can't be sure what its failing on is what you're expecting it to fail on, just that one of the fails has been triggered

packages/gp_file_parser/file_name_parser.py
abbas-khan8: Valid point, the `_` was left in the regex accidentally. I agree with the 1 returned error message for things like invalid format or incorrect characters, but I still think multiple messages for individual errors is required. If the filename contained an invalid practice code and invalid HA cipher, they should be made aware of both right? Otherwise they'd correct the practice code and upload again, only to be informed that the HA cipher needed fixing too

test/unittests/packages/gp_file_parser/test_filename_parser.py
```diff
@@ -33,54 +33,149 @@ def test_validate_filename_valid_lowercase_filename_returns_valid_date():
...
+@freeze_time("2020-02-01")
+def test_validate_filename_invalid_format_raises_InvalidFilename():
+    filename = "A12023GPR4BRF1.B1A"
+
+    expected = "Filename does not have the correct format"
+
+    with pytest.raises(InvalidFilename) as exc:
+        validate_filename(filename)
+
+    assert str(exc.value) == expected
```
josh-scargill-nhsd, line 54: Does this not work without `sty()`?

test/unittests/packages/gp_file_parser/test_filename_parser.py
```diff
@@ -33,54 +33,149 @@ def test_validate_filename_valid_lowercase_filename_returns_valid_date():
...
+@freeze_time("2020-02-01")
+def test_validate_filename_invalid_characters_raises_InvalidFilename():
+    filename = "A12023-GPR4BRF1.B1A"
+
+    expected = "Filename contains invalid characters"
+
+    with pytest.raises(InvalidFilename) as exc:
+        validate_filename(filename)
+
+    assert str(exc.value) == expected
```
josh-scargill-nhsd, line 66: Is there any value in doing a `pytest.mark.parametrize` here?

josh-scargill-nhsd (approved):

test/unittests/packages/gp_file_parser/test_filename_parser.py
abbas-khan8: No it seems to return the exception as well as the message, adding `str()` just gets the message

test/unittests/packages/gp_file_parser/test_filename_parser.py
abbas-khan8: There was a reason why I didn't, it might have been due to my incorrect approach in terms of syntax but I think it could have been due to using `freezetime` while also using `parametrize`. Unless you see any value in it?

jamesnunnnhs (approved):

PR #25, opened 2021-06-29
PCRM-3845 - Create consolidated file for demographic differences

jamesnunnnhs:
Creates one or more work items in a bucket, one per patient for a job, where one or more demographics differences has been identified for a patient.


lambdas/LR_15_process_demo_diffs/process_demographic_differences.py
```diff
@@ -153,10 +195,13 @@ def process_demographic_differences(job_id: str) -> Dict:
 
     out_files = []
 
+    summary_records = []
+
+    s3 = boto3.client("s3")
```
joeds13, line 200: I think we want to initialise this `S3 client` at the module level, not inline in this function

I'm struggling to find the documentation to point you to, but when it's defined at the module level it is initialised when the lambda first starts and will not get reinitialised on every run like it will being in this function

The boto3 library transparently handles ensuring you have credentials/a session when it's called so there should be no issues when using the client here, but we will see potentially seconds worth of time saved on a warm lambda request due to boto not repeatedly initialising

jamesnunnnhs: Ok fair enough I didn't know that, but what are the consequences of doing this? We have several instances of this too so need to consider refactoring if it'll be a problem @josh-scargill-nhsd?

pytest.ini
```diff
@@ -1,5 +1,9 @@
 [pytest]
 env =
+    AWS_ACCESS_KEY_ID=testing
+    AWS_SECRET_ACCESS_KEY=testing
+    AWS_SECURITY_TOKEN=testing
+    AWS_SESSION_TOKEN=testing
```
josh-scargill-nhsd, line 6: I was speaking with Abbas earlier, why have you done this? I don't think we should be specifying these anywhere

jamesnunnnhs: This is for running tests on github only, hence being in pytest.ini. When I changed where the client is defined in response to what Joe said, the client is set up as soon as the module is imported which meant that when running for tests, boto3 expects some proper credentials defined on github. I couldn't mock them with moto because the imports appear to occur before moto sets up, therefore boto3 complains about not having credentials.

aparimoo-nhs: this could be done in `conftest` ( at global level) using [monkeypatch](https://docs.pytest.org/en/6.2.x/monkeypatch.html)  by doing
monkeypatch.setenv("AWS_ACCESS_KEY_ID","testing") , no need to do in `.ini` file

josh-scargill-nhsd (approved):

lambdas/LR_15_process_demo_diffs/process_demographic_differences.py
josh-scargill-nhsd: If we know how to do that with the new template then great, if not lets just push on for now and stick it as a refactor piece

abbas-khan8 (approved):

PR #28, opened 2021-06-30
PCRM-3781 New Terraform Stacks and fmt

joeds13:
Create the new List Reconciliation terraform stack, this will act as a
single source of terraform that will be used across all environments,
this will stop the duplication of code under the environments directory
and help to ensure environment consistency
We have also run a terraform fmt to ensure all code is formatted
consistently, we will later add terraform fmt check and validation to
the CI process

Running fmt ensures consistent code formating for terraform, so we will
catch it here and expect code to be formatted pre-commit, later we will
add a precommit hook to make this easier
Running validation does a static terraform check, again should be done
precommit, giving a faster feedback loop and potentially catching some
bad terraform pre-applying that may result in an inconsistent aws
environment

Allows us to control if test data is deployed to an environment from a
single location


jamesnunnnhs (approved):

terraform/stacks/list-reconciliation/outputs.tf
```diff
@@ -0,0 +1,110 @@
...
+}
+
+output "jobs_Status_table" {
+  value = module.list-rec.Jobs_Status_Table
+}
+
+output "statuses_table" {
+  value = module.list-rec.Statuses_Table
+}
+output "LR-12-lambda" {
```
aparimoo-nhs, line 96: new line here

terraform/modules/test-data/s3.tf
```diff
@@ -0,0 +1,23 @@
+resource "aws_s3_bucket_object" "upload-mock-pds-data" {
```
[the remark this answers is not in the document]
joeds13: Yeah, depending on the feedback to "Consolidate multiple resource outputs to single object output" in https://nhsd-confluence.digital.nhs.uk/display/PCRM/PCRM-3718+Ticket+notes we will tackle that

Then also when/if we adopt this stack over the existing environments, we'll need to see the integration tests pass so will ensure they're joined up correctly

aparimoo-nhs (approved):

makefile
```diff
@@ -3,6 +3,10 @@ terraform_workspace ?= default
 
 all: python-package init apply
 test: integrationtests unittest
+format: fmt black
```
josh-scargill-nhsd, line 6: Just to make everyones life easier, could we have a make command for pre commit? just a single 1 stop command i can run before commiting? or does tf take time so you've not combined it?

makefile
```diff
@@ -3,6 +3,10 @@ terraform_workspace ?= default
...
+
+get-branch-id:
```
josh-scargill-nhsd, line 8: I'm also wondering, this makefile is getting a bit messy, assuming we can only have a single makefile, can we comment into a few different sections or something?

makefile
joeds13: Is there anything else we want to do alongside the formatting precommit?

I'm hoping that the hooks will be able to do everything for us, I've made a start here: https://github.com/NHSDigital/list-reconciliation/compare/PCRM-3781-code-formatting-git-hooks

makefile
joeds13: I think this Makefile is pretty clear at the minute, it is less than a single screen in my editor, and I'm personally not a massive fan of comments breaking up sections, but happy to be guided?

makefile
josh-scargill-nhsd: That's all gravy then 👍

makefile
josh-scargill-nhsd: Theres a lot of commands. I think at minimum splitting between formatting and TF. But think we should break the TF down so testing has it's own section, can't see what else stands out without that in place, 1 worded comments are far quicker to read than multiple commands imo :)

joeds13: How's this for you?
https://github.com/NHSDigital/list-reconciliation/pull/28/commits/6667fc7e53109aa0ba9e2a169bb0811246f38d9a

josh-scargill-nhsd: This looks better I think, we can trial it and see how we find it :)

josh-scargill-nhsd (approved):

PR #35, opened 2021-07-05
PCRM-3781 PreProd Deployment

joeds13:
Selects the correct deployment role when in the PreProd workspace
POCing separating jobs within a Github Action

We are using hardcoded `terraform` commands, rather than make commands, as we are still transitioning between `terraform/environments/*` and `terraform/stacks/list-reconciliation` and have not updated the paths in the `makefile` yet, which we will do as part of the PR to remove `terraform/environments/*`


terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -3,6 +3,12 @@ locals {
   environment        = terraform.workspace
   data_clasification = local.environment == "prod" ? "5" : "1"
 
+  terraform_deploy_role_arn = {
+    dev     = "arn:aws:iam::092420156801:role/LRTerraformDeploy"
+    preprod = "arn:aws:iam::287634746327:role/LRTerraformDeploy"
+    # prod = "arn:aws:iam::000000000000:role/LRTerraformDeploy"
```
aparimoo-nhs, line 9: remove comment

aparimoo-nhs (approved):

josh-scargill-nhsd (approved):

PR #36, opened 2021-07-07
Restored missing lambda event

aparimoo-nhs:
Fixed LR-15 trigger where event was missing for job_id


jamesnunnnhs (approved):

premnhs (approved):
LGTM

packages/lambda_code/LR_15_process_demo_diffs/lr15_lambda_handler.py
```diff
@@ -30,6 +30,7 @@ def initialise(self):
     def start(self):
 
         try:
+            self.job_id = self.event["job_id"]
```
josh-scargill-nhsd, line 33: Is this the right place to do it? I thought the whole point of our class was to get this kind of thing?
In LR 11 we're doing it in initialise for example

PR #40, opened 2021-07-09
Fix passing of terraform outputs for tests in the PreProd Env

joeds13:
Once we have more of terraform refactor in place, these Workflows will move back to using the Makefile as the source of truth for commands and we won't see issues like this


premnhs (approved):
LGTM

aparimoo-nhs (approved):

PR #41, opened 2021-07-12
PCRM-3719 Adding Pre-Commit

joeds13:
Pre-Commit to manage githooks with some standard checks and List Rec
specific checks


joeds13:
When this is merged, I'll add instructions here (which has superseded the README) with the required instructions for installing the hooks:
- After `pip install -r test_requirements.txt`
- run `make hooks`

test_requirements.txt
```diff
@@ -8,4 +8,5 @@ getgauge
 moto
 pytest-env
 docker
```
josh-scargill-nhsd, line 10: Should we have versions here too?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-yaml
+      - id: check-added-large-files
+      - id: check-case-conflict
+      - id: check-merge-conflict
```
josh-scargill-nhsd, line 12: Just for information to others:
```
trailing-whitespace - This hook trims trailing whitespace. 
end-of-file-fixer - Ensures that a file is either empty, or ends with one newline. 
check-yaml - This hook checks yaml files for parseable syntax. 
check-added-large-files - Prevent giant files from being committed (default = 500kb)
check-case-conflict - Check for files that would conflict in case-insensitive filesystems 
check-merge-conflict - Check for files that contain merge conflict strings. 
```

test_requirements.txt
joeds13: It's generally good practise to at least have the minimum required versions, so I did that for deps affected by this PR, we pin the version of black as we pin it in the pre-commit config
But getting the minimum versions of all existing test deps is definitely outside of the scope of this PR

.pre-commit-config.yaml
joeds13: Once this is merged, we can do a follow up PR with `pre-commit run --all-files` to get us up to a good baseline

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
```
jamesnunnnhs, line 3: Taking suggestions? 

https://github.com/aquasecurity/tfsec
https://github.com/PyCQA/bandit
https://github.com/timothycrosley/isort

Only because I've used them before, maybe worth using maybe not.

jamesnunnnhs (approved):

.pre-commit-config.yaml
joeds13: We've just started discussing some of the security stuff, we used `tfsec` to give us a baseline, so when we remediate the initial findings from that adding it in here would no doubt be a good call to warn of regressions. Not seen Bandit but we could also look to use that too.

As for isort think we probably want a bit of a team decision, there's also flake8 etc that are in the Github Workflows, do we just want everything in there? Maybe one to raise in the Team chat tomorrow and we can raise future PRs to include what we decide?

aparimoo-nhs (approved):

.pre-commit-config.yaml
josh-scargill-nhsd: 👍 for sorting imports, definitely something we're missing

josh-scargill-nhsd (approved):

PR #46, opened 2021-07-13
PCRM-3731 - Solve file splitting performance issue

jamesnunnnhs:
Replaces SQS queue with lambda to split file into S3 objects.


packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
```diff
@@ -183,22 +193,35 @@ def handle_validated_records(self, validated_file: dict) -> success:
...
+
+            lambda_.invoke(
+                FunctionName=self.system_config["LR_24_SAVE_RECORDS_TO_S3"],
+                InvocationType="Event",
+                Payload=json.dumps(body),
+            )
+
     def write_to_dynamodb(self, validated_file: dict, num_of_records: int) -> list:
         """Creates Job items and writes a batch of validated records to DynamoDb.
             Appends 'Id' field to each validated patient record in records
```
abbas-khan8, line 213: docstring needs updating to match new purpose?

packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
```diff
@@ -183,22 +193,35 @@ def handle_validated_records(self, validated_file: dict) -> success:
...
-                HA Cipher and a list of valid patient records
+            practice_code (str): HA cypher of GP extract
+            upload_filename (str): Filename of uploaded GP extract
             num_of_records (int): Number of validated records
 
         Returns:
             Records (list): List of Records with added ID field
         """
 
         practice_code = validated_file["practice_code"]
```
abbas-khan8, line 224: Since a chunk of the original functionality to do with `validated_file` dict has been removed, maybe this function just needs `practice_code` string?

packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
```diff
@@ -141,7 +150,7 @@ def handle_validated_records(self, validated_file: dict) -> success:
         )
 
         try:
-            records = self.write_to_dynamodb(validated_file, num_of_records)
+            self.write_to_dynamodb(validated_file, num_of_records)
```
abbas-khan8, line 153: Since a chunk of the original functionality to do with validated_file dict has been removed from `write_to_dynamodb`, maybe this function just needs practice_code string as a parameter?

terraform/modules/lambda/LR-07/iam.tf
```diff
@@ -1,5 +1,5 @@
 locals {
-  patient_sqs_queue = "arn:aws:sqs:::${var.patient_sqs}"
+  lr-06-bucket_arn = "arn:aws:s3:::${var.lr-06-bucket}"
```
joeds13, line 2: We should pass the ARN in here, rather than constructing it in a string

terraform/modules/s3/outputs.tf
```diff
@@ -44,5 +44,9 @@ output "buckets" {
       bucket = aws_s3_bucket.LR-23.id
       arn    = aws_s3_bucket.LR-23.arn
     }
+    LR-06 = {
```
joeds13, line 47: Move to line 31

Keep the numerical ordering of buckets?

terraform/stacks/list-reconciliation/modules.tf
```diff
@@ -81,10 +65,8 @@ module "lr_10_registration_orchestration" {
 
 module "test_data" {
   # only load test data in the non-prod accounts
-  count = local.environment != "prod" ? 1 : 0
-
-  source = "../../modules/test_data"
-
+  count        = local.environment != "prod" ? 1 : 0
```
joeds13, line 68: I like to add spacing here, as it separates the module meta arguments "source"/"count" from the actual module variables "suffix"/"LR_22_bucket"/etc just makes it a bit clearer to read later

joeds13 (changes requested):

packages/lambda_code/LR_24_save_records_to_s3/lr24_lambda_handler.py
```diff
@@ -0,0 +1,97 @@
...
+from utils.logger import success
+from gp_file_parser.utils import empty_string
+
+S3 = boto3.client("s3")
+
+cwd = os.path.dirname(__file__)
+ADDITIONAL_LOG_FILE = os.path.join(cwd, "..", "..", "utils/cloudlogbase.cfg")
+
+
+class SaveRecordsToS3(LambdaApplication):
```
josh-scargill-nhsd, line 21: I thought we discussed that this was to be a generic bucket saving mechanism?

packages/utils/database/models.py
```diff
@@ -70,8 +70,8 @@ class Demographics(Model):
     class Meta:
         table_name = DEMOGRAPHICS_TABLE
         region = AWS_REGION
-        read_capacity_units = 1
-        write_capacity_units = 1
+        read_capacity_units = 1000
+        write_capacity_units = 1000
```
josh-scargill-nhsd, line 74: Can we do this on others if they're on 1 too? in case it does affect, I don't think we want 1 anywhere

terraform/modules/lambda/LR-08/iam.tf
```diff
@@ -74,9 +74,18 @@ resource "aws_iam_policy" "policy" {
...
             "Resource": [
-                "${var.demographics_differences_table_arn}",
                 "${var.errors_table_arn}"
             ]
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "dynamodb:BatchWriteItem",
+                "dynamodb:PutItem"
```
josh-scargill-nhsd, line 84: Doesn't it just need 1 of these?

packages/utils/database/models.py
jamesnunnnhs: Er yeah I can change the rest, but I really don't know where that one came from, I didn't touch these

joeds13: Do we want to define these in Terraform, rather than in the code?

We may want them to differ per env, if they are provisioned capacity units they incur standing charges that we won't want in Dev

terraform/modules/lambda/LR-08/iam.tf
jamesnunnnhs: yes

packages/utils/database/models.py
jamesnunnnhs: I still think it's arbitrary-  the rest are set to 100, keep the same? or all set to 1000? or 100?

jamesnunnnhs: Thing is @joeds13 that these are needed for the python lib to work but don't actually do anything in practice except if we use the python lib to provision the tables, which we don't. They do nothing in this case which is why they are set to 1 on the other tables, I don't believe there's any throttling or retrying based on these values

jamesnunnnhs: So yes they should be set in terraform if it's moved from on demand to provisioned.

packages/lambda_code/LR_24_save_records_to_s3/lr24_lambda_handler.py
jamesnunnnhs: you're right, I forgot that bit

packages/utils/database/models.py
josh-scargill-nhsd: Since we don't know whats going on under the hood, I think making them all 1000 is a safe bet to avoid any potentially throttling that we aren't anticipating

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -16,49 +19,84 @@
...
+            .decode("utf-8")
+        )
 
-                sqs.delete_message(
-                    QueueUrl=sqs.get_queue_url(
-                        QueueName=queue_name, QueueOwnerAWSAccountId=account_id
-                    )["QueueUrl"],
-                    ReceiptHandle=record["receiptHandle"],
-                )
+        job_id = body["job_id"]
```
josh-scargill-nhsd, line 37: Why is this created here instead of in the patient?

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -16,49 +19,84 @@
...
             raise
 
         except Exception as err:
-            msg = f"Unhandled error patient_id: {patient_id} nhs_number: {nhs_number}"
+            msg = f"Unhandled error JobId: {job_id}, PatientId: {patient.Id} NhsNumber: {patient.NhsNumber}"
             error_response = log_dynamodb_error(self.log_object, job_id, UNHANDLED_ERROR, msg)
 
-            raise type(err)(error_response) from err
+            raise Exception(error_response) from err
+        else:
```
josh-scargill-nhsd, line 75: Since all the exceptions raise an error, is this `else` required? code the code below not be part of the main body?

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -16,49 +19,84 @@
...
+            GP_PostCode=empty_string(body["POSTCODE"]),
+            GP_DrugsDispensedMarker=str(body["DRUGS_DISPENSED_MARKER"]),
+        )
 
-            self.response = json.dumps(self.pds_hydrate(nhs_number, job_id, patient_id))
+        try:
+            json.dumps(self.pds_hydrate(job_id, patient))
 
         except PDSAPIError:
             raise
```
josh-scargill-nhsd, line 68: I'm not sure if this is needed? In the event `PDSAPIError` is raised, it's going to stop the application been processed no?

packages/lambda_code/LR_24_save_records_to_s3/lr24_lambda_handler.py
```diff
@@ -0,0 +1,66 @@
...
+                "logger": "LR24.Lambda",
+                "level": "INFO",
+                "message": f"Splitting {source} {len(records)} records",
+            },
+        )
+        for record in records:
+            record_dict = json.loads(record)
+            id_ = "_".join([str(record_dict[col]) for col in id_cols])
+            try:
+
```
josh-scargill-nhsd, line 41: Can we move `try` down a line, so it has a new line above, and no new line below

packages/lambda_code/LR_24_save_records_to_s3/lr24_lambda_handler.py
```diff
@@ -0,0 +1,66 @@
...
+                self.retry_put_object(
+                    destination_bucket,
+                    f"{id_}/{id_}.json",
+                    json.dumps(record_dict),
+                )
+            except Exception as err:
```
josh-scargill-nhsd, line 47: new line before except

packages/lambda_code/LR_24_save_records_to_s3/lr24_lambda_handler.py
```diff
@@ -0,0 +1,66 @@
...
+            },
+        )
+        for record in records:
+            record_dict = json.loads(record)
+            id_ = "_".join([str(record_dict[col]) for col in id_cols])
+            try:
+
+                self.retry_put_object(
+                    destination_bucket,
+                    f"{id_}/{id_}.json",
```
josh-scargill-nhsd, line 44: Not sure we want it like this do we? aka: `myUniqueId/myUniqueId.json`, seems a bit unusual 🤔

packages/services/split_records_to_s3.py
```diff
@@ -0,0 +1,57 @@
...
+# 128kb max Event payload size, minus enough to allow for job_id and practice_code in the body
+PAYLOAD_SIZE = (128 * 1024) - 200
+
+
+def split_records_to_s3(
+    records: List[str],
+    id_cols: List[str],
+    destination_bucket: str,
+    source: str,
+    max_size: int = 128 * 1024,
```
josh-scargill-nhsd, line 19: Not sure if we want to make this available, couldn't the caller pass in too high a value and cause it to fail?

joeds13 (approved):

test/unittests/lambdas/LR_24_save_records_to_s3/test_LR_24_save_records_to_s3.py
```diff
@@ -0,0 +1,132 @@
...
+                    "ADDRESS_LINE4": "",
+                    "ADDRESS_LINE5": "",
+                    "POSTCODE": "HP14 2YQ",
+                    "DRUGS_DISPENSED_MARKER": False,
+                }
+            ),
+            json.dumps(
+                {
+                    "job_id": "789",
+                    "practice_code": "Y12345",
```
josh-scargill-nhsd, line 47: Since these 2 objects are going to be the same, can we store them as variables instead? just shortens the test to make it a bit more readable :)

test/unittests/lambdas/LR_24_save_records_to_s3/test_LR_24_save_records_to_s3.py
```diff
@@ -0,0 +1,132 @@
...
+import boto3
+
+ROOT = os.path.dirname(os.path.abspath(__file__))
+DATA = os.path.join(ROOT, "..", "data")
+
+LR_06_BUCKET = os.environ.get("LR_06_BUCKET")
+REGION_NAME = os.environ.get("AWS_REGION")
+
+
+def test_save_records_to_s3_saves_records(
```
josh-scargill-nhsd, line 12: I'd expect a test for the exception been thrown as expected. And the retrying logic if we aren't testing it elsewhere?

packages/lambda_code/LR_24_save_records_to_s3/lr24_lambda_handler.py
jamesnunnnhs: It makes no difference to the logic but it does get round a limit I was hitting where you can only have so many requests per second under a single prefix (i.e. s3 folder). Creating a new prefix for each file means unlimited rps

packages/services/split_records_to_s3.py
jamesnunnnhs: they could want set it lower

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
jamesnunnnhs: because job_id is used in loads of places after this point

josh-scargill-nhsd: but we could just use patient.JobId instead if we set that since it's set directly below no?

josh-scargill-nhsd (approved):

abbas-khan8 (approved):

PR #47, opened 2021-07-14
PCRM-3515 -  Feedback GP extract validation errors

abbas-khan8:
- Triggers on LOG file creation in s3 `fail/logs/` folder
- Reads and validates LOG file, and constructs email message from contents
- Deletes failed file


terraform/modules/lambda/lambda.tf
```diff
@@ -30,6 +31,21 @@ module "LR-02" {
...
+  source            = "./LR-04"
+  lambda_name       = local.lambda_name.LR-04
+  package_layer_arn = aws_lambda_layer_version.package_layer.arn
+  runtime           = var.runtime
+  source_bucket     = var.s3_buckets.LR-01.bucket
+  errors_table_arn  = var.dynamodb_tables.errors.arn
+  errors_table_name = var.dynamodb_tables.errors.name
+  suffix            = var.suffix
+  lambda_handler    = var.lambda_handler
+  depends_on        = [module.LR-02.LR_02_lambda]
```
joeds13, line 48: We shouldn't need to use this to create a dependency

If it is needed, we should be passing through some resources/data from that module, then when terraform creates it's DAG it will be able to see the dependency itself

joeds13 (changes requested):

packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
```diff
@@ -335,35 +342,40 @@ def handle_extract(self, prefix: str, error_message: str = None):
...
+    def process_invalid_message(self, exception: Exception) -> str:
         """Create a formatted error message string based on raised
             exception
 
         Args:
             exception (Exception): exception raised
 
         Returns:
-            str: Error message as formatted string.
+            str: JSON dump of error log dict containing Job ID, practice code, filename and error message/s.
```
jamesnunnnhs, line 358: I'd recommend just returning a dict, then converting to json as late as possible, just makes it easier to read what you need from the object until you dump as text.

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+            )
+
+            raise FeedbackLogError(error_response) from err
+
+        self.failed_key = FAILED_PREFIX + self.log["file"]
+
+    def send_email(self):
+        """Send an email based on LOG info"""
+
+        to = "test@nhs.net"
```
jamesnunnnhs, line 198: This needs to be an environment variable I believe, and it will need to be set per terraform environment

terraform/modules/lambda/LR-02/lambda.tf
```diff
@@ -28,7 +28,7 @@ resource "aws_lambda_function" "LR-02-Lambda" {
       AWS_S3_REGISTRATION_EXTRACT_BUCKET = var.source_bucket
       DEMOGRAPHICS_TABLE                 = var.demographics_table_name
       JOBS_TABLE                         = var.jobs_table_name
-      INFLIGHT_TABLE                     = var.in_flight_table_name
+      INFLIGHT_TABLE                     = var.in_flight_table_arn
```
jamesnunnnhs, line 31: Is it correct to change this? The arn works with pynamodb does it?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+        """
+
+        self.read_log()
+        self.log_object.write_log(
+            "UTI9995",
+            None,
+            {
+                "logger": "LR04.Lambda",
+                "level": "INFO",
+                "message": "LR-04 successfully read and validated log file",
```
josh-scargill-nhsd, line 106: For which job?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+
+        self.send_email()
+        self.log_object.write_log(
+            "UTI9995",
+            None,
+            {
+                "logger": "LR04.Lambda",
+                "level": "INFO",
+                "message": "LR-04 successfully sent email",
```
josh-scargill-nhsd, line 117: For which job?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+        self.cleanup_files()
+        self.log_object.write_log(
+            "UTI9995",
+            None,
+            {
+                "logger": "LR04.Lambda",
+                "level": "INFO",
+                "message": "LR-04 successfully cleaned up S3 files",
+            },
+        )
```
josh-scargill-nhsd, line 130: Are we wanting to remove log files? Is the process that a job fails we're referring purely to splunk to see what went wrong? as I don't think the information in their is substantial enough

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+                "99999999-04-0404-0404-999999999999",
+                "HANDLED_ERROR",
+                msg,
+            )
+
+            raise FeedbackLogError(error_response) from err
+
+        self.log = log
+
+        self.validate_log()
```
josh-scargill-nhsd, line 161: Looking at how you set your code up, i'd suggest this call was in the parent function

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+    INVALID_STRUCTURE,
+)
+
+from utils.logger import log_dynamodb_error, success
+from utils.exceptions import FeedbackLogError
+
+cwd = os.path.dirname(__file__)
+ADDITIONAL_LOG_FILE = os.path.join(cwd, "..", "..", "utils/cloudlogbase.cfg")
+
+FAILED_PREFIX = "fail/"
```
josh-scargill-nhsd, line 23: This is used in multiple places, I question that we should be storing this somewhere more generic now

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+    def validate_log(self):
+        """Validates LOG dictionary's structure and data values. If successful, set the `failed_key`
+
+        Raises:
+            FeedbackLogError: If validation fails
+        """
+
+        try:
+            log_name = str(self.log_key).replace(FAILED_PREFIX, "")
+            assert f"{self.log['file']}_LOG.json" == log_name
```
josh-scargill-nhsd, line 172: We've discussed before not to use `assert` in none test code

packages/gp_file_parser/file_name_parser.py
```diff
@@ -27,48 +27,43 @@ def validate_filename(filename: str, process_datetime: datetime = None) -> dict:
     filename = filename.upper()
 
     if len(filename) > 19:
-        msg = "Filename exceeds character limit"
-
-        raise InvalidFilename(msg)
+        msg = ["Filename exceeds character limit"]
```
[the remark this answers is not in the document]
abbas-khan8: It's for consistency really, some error messages (filename errors/invalid records) can contain multiple errors so returning error inside a list here so the structure of the log looks the same

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+                msg,
+            )
+
+            raise FeedbackLogError(error_response) from err
+
+        self.log = log
+
+        self.validate_log()
+
+    def validate_log(self):
```
josh-scargill-nhsd, line 163: This function seems a bit strange.

If you're validating if things are ok, maybe go with a more standard approach of `if something bad, raise something`?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
abbas-khan8: This is a stub atm, it will be expanded and improved in a later PR. That being said the implementation code they provided us with don't use env variables for any of the email components

terraform/modules/lambda/LR-02/lambda.tf
abbas-khan8: Nope I'm not 100% why this has changed tbh, it's a mistake

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+
+            invalid_ids = 0
+            invalid_records = 0
+            invalid_records_msg = ""
+
+            for record in records:
+                if "TRANS_ID" in record["_INVALID_"]:
+                    invalid_ids += 1
+
+                if record["TRANS_ID"] != "None":
```
josh-scargill-nhsd, line 236: I don't see an else, wasn't this the line number conversation?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -0,0 +1,295 @@
...
+            "UTI9995",
+            None,
+            {
+                "logger": "LR04.Lambda",
+                "level": "INFO",
+                "message": f"{to}\n{subject}\n{body}",
+            },
+        )
+
+    def create_body(self) -> str:
```
josh-scargill-nhsd, line 212: Looking at this, it doesn't seem to match the spec noted here https://nhsd-confluence.digital.nhs.uk/display/PCRM/PCRM-3515+%28PDS-299%29+-+Identify+any+data+validation+errors

josh-scargill-nhsd (commented):
I assume testers are happy you've noodled with their tests?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
abbas-khan8: Where would the line number come from? The parser would have to be changed to store line numbers wouldn't it?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
abbas-khan8: There is not Job, a Job item is only created if validation passed

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
abbas-khan8: I don't have a preference, I'm not against keeping the log files. At this point there is no job, once an email has been sent back to PCSE informing them of the failed file the process is complete right? The log file never has to be mentioned or used again, unless we want to retain it as proof?

terraform/modules/lambda/LR-02/lambda.tf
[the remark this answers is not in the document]
abbas-khan8: This was a mistake, I've reverted changes in latest push

packages/gp_file_parser/file_name_parser.py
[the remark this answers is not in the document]
jamesnunnnhs: So you try to parse what you can and return multiple errors if there are multiple with just the filename?

test/integrationtests/specs/lr_02_lambda_others.spec
```diff
@@ -4,6 +4,7 @@
 
 ## test to ensure when lambda LR-02 is triggered and check file is validated and passed sucessfully
 ---------------------------------------------------------------------------------------------------
+Tags: wip
```
HarveyNHS, line 7: Hey Abbas, I've noticed you've added wips tags to these tests, are you checking for something within the Githubs pipeline?

HarveyNHS (commented):
Added review

test/integrationtests/specs/lr_02_lambda_others.spec
abbas-khan8: Hi Harvey, these have been marked WIP due to a convo had with Prem. Because the LOG file was getting deleted from the fail/ folder once consumed in the process, the integration tests which were checking for the existence/validity of the LOG file were failing as the file had been deleted

packages/gp_file_parser/file_name_parser.py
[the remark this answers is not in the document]
abbas-khan8: LR04 will treats any filename errors as a list when it comes to processing these messages, because there's a chance this will return more than one error message. I get this might not be the most efficient way but the rewriting/reformatting required to change this would result in very little performance increase, is it worth it or is there another issue that I'm not seeing here?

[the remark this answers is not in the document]
jamesnunnnhs: nope just wondering

joeds13 (approved):

jamesnunnnhs (approved):

PR #48, opened 2021-07-15
PCRM-3989 - Adds logging and versioning to buckets

jamesnunnnhs:


terraform/modules/s3/LR-01.tf
```diff
@@ -12,6 +12,15 @@ resource "aws_s3_bucket" "LR-01" {
...
     Name = "S3 Input Bucket for LR-01 - ${var.suffix}"
   }
+
+  versioning {
+    enabled = true
+  }
+
+  logging {
+    target_bucket = aws_s3_bucket.LR-26.id
+    target_prefix = "log/"
```
joeds13, line 22: Might be better to namespace the logs on each bucket:
`target_prefix = "lr-01-gp-extract-input-${var.suffix}/"`

As the files just turn up as `s3://list-rec-access-logs-jamesnunn/log/2021-07-15-13-13-03-329E618F5B874B67`

joeds13 (changes requested):

joeds13 (approved):

premnhs (approved):
LGTM

aparimoo-nhs (approved):

PR #49, opened 2021-07-15
PCRM-3538 LR-25 MESH Post Office Lambda

joeds13:
When open, delivers messages from Mesh inbound directories to their
intended List Reconciliation targets


packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,103 @@
...
+                "UTI9995",
+                None,
+                {
+                    "logger": "LR25.Lambda",
+                    "level": "INFO",
+                    "message": "post office status=closed'",
+                },
+            )
+            self.response = '{"post_office": "closed"}'
+            return
```
aparimoo-nhs, line 28: I think `return ` is not needed , `self.response` is suffice

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,103 @@
...
+                "logger": "LR25.Lambda",
+                "level": "INFO",
+                "message": "post office status=open'",
+            },
+        )
+        self.response = '{"post_office": "open"}'
+        return
+
+    def get_mesh_messages(self):
+        mesh_bucket_id = os.environ.get("MESH_BUCKET_ID")
```
aparimoo-nhs, line 43: this should be `self.system_config['MESH_BUCKET_ID']`

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,103 @@
...
+                    "message": f"found message={message}",
+                },
+            )
+        return messages
+
+    def check_if_open(self) -> bool:
+        """
+        Check to see if the Post Office is open
+        """
+        if os.environ.get("POST_OFFICE_OPEN") == "True":
```
aparimoo-nhs, line 61: should be `self.system_config["POST_OFFICE_OPEN"]`

terraform/modules/lambda/lambda.tf
```diff
@@ -168,3 +169,18 @@ module "LR-21" {
   suffix                      = var.suffix
   lambda_handler              = var.lambda_handler
 }
+
+module "lr_25" {
+  source = "./LR_25_mesh_post_office"
+
+  # TODO remove this overwrite
+  mesh_post_office_open = "True"
+  # mesh_post_office_open = var.mesh_post_office_open
```
aparimoo-nhs, line 178: do you need this comment ?

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,103 @@
...
+            return True
+        return False
+
+    def s3_move_file(self, bucket, old_key, new_key, delete_original=True) -> None:
+        """
+        Move a file between locations in S3
+        """
+        s3r = boto3.resource("s3")
```
aparimoo-nhs, line 69: consider using  exception catching

Makefile
```diff
@@ -26,7 +26,7 @@ python-package:
 
 # Testing
 unittests:
-	pytest --ignore=test/integrationtests
+	pytest
```
aparimoo-nhs, line 29: why removed ignored? this will start discovering `integrationtests ` which will fail

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,103 @@
...
+        self.log_object.write_log(
+            "UTI9995",
+            None,
+            {
+                "logger": "LR25.Lambda",
+                "level": "INFO",
+                "message": "post office status=open'",
+            },
+        )
+        self.response = '{"post_office": "open"}'
```
aparimoo-nhs, line 39: I think `return ` is not needed , `self.response` is suffice

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
joeds13: https://github.com/NHSDigital/spine-core-aws-common#usage

Pattern taken from the docs

Makefile
joeds13: https://github.com/NHSDigital/list-reconciliation/blob/develop/pytest.ini#L37

You already define the ignore in the `pytest.ini` we don't need the duplication of the cli flag

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
joeds13: Again, taken from https://github.com/NHSDigital/spine-core-aws-common#usage

joeds13:
Blocked my https://github.com/NHSDigital/spine-core-aws-common/pull/33

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,133 @@
...
+            self.log_object.write_log(
+                "UTI9995",
+                None,
+                {
+                    "logger": "LR25.Lambda",
+                    "level": "INFO",
+                    "message": "post office status=closed'",
+                },
+            )
+            self.response = '{"post_office": "closed"}'
```
jamesnunnnhs, line 30: is it worth reusing the same message in the response?
```
msg = 'post office status closed'
{    ...
     "message": msg,
    ...
}
self.response = json.dumps({"status": msg})
```
Also personally I try to not define literal json - instead define literal dicts and then `json.dumps` as it helps prevent any malformed handwritten json.

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        List all files in an S3 Bucket
+        """
+        s3r = boto3.resource("s3")
+        bucket = s3r.Bucket(bucket_name)
+        all_objects = bucket.objects.all()
+        objects = []
+        for object in all_objects:
+            if object.key.startswith(prefix) and not object.key.endswith("/"):
+                objects.append(object.key)
```
jamesnunnnhs, line 132: Use this with the `Prefix` arg to avoid your own loop?https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.list_objects_v2

terraform/modules/lambda/LR_25_mesh_post_office/iam.tf
```diff
@@ -0,0 +1,132 @@
...
+      sid    = "S3List${statement.value.name}"
+      effect = "Allow"
+
+      actions = [
+        "s3:ListBucket",
+      ]
+
+      resources = [
+        "arn:aws:s3:::${statement.value.inbound.bucket}",
+        "arn:aws:s3:::${statement.value.outbound.bucket}"
```
jamesnunnnhs, line 75: I think you asked me previously to use a predefined arn rather than construct a bucket arn from the name ;-)

jamesnunnnhs: ditto the ones below

test/unittests/lambdas/LR_25_mesh_post_office/test_LR_25_mesh_post_office.py
```diff
@@ -0,0 +1,72 @@
...
+#   The conflict is due to cacheing of SSM Parmeters in the spine_aws_common library
+#   once we are on a library version that contains this fix, we can add this test back in:
+#   https://github.com/NHSDigital/spine-core-aws-common/pull/33
+# def test_check_if_open_should_be_closed(mesh_post_office_closed):
+#     """
+#     If the mesh_post_office_open parameter is falsey, we should expect the Post Office to be Closed
+#     """
+#     response = mesh_post_office_closed.check_if_open()
+#     expected = False
+#     assert response == expected
```
jamesnunnnhs, line 29: If you want to keep the code, uncomment and just mark it with `@pytest.mark.skip('conflict due to caching of SSM Parameters in the spine_aws_common library')`

jamesnunnnhs (changes requested):

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
joeds13: Yeah I thought about this, but I couldn't find a way to list with both the prefix and suffix check, as we want to ignore "directories" so never rewrote this one

Still not sure if there's a cleaner way to do both prefix/suffix on list and this piece of code is not very performance sensitive, so not sure how much effort it's worth looking into?

terraform/modules/lambda/LR_25_mesh_post_office/iam.tf
joeds13: Yeah, I was expecting this one...

Issue here is that we need access to an ARN which is unknown, the var's (`statement.value.*.bucket`) get passed in as config to the module, so we could just require ARNs to be passed in there, but that just moves the problem

We could also use the `aws_s3_bucket` data resource, to lookup the ARN via the name, but that isn't so clean or easy to pass into this loop, with the language functions that terraform provides, so I'm not really a fan of that either

Happy to be challenged further

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
jamesnunnnhs: I'm not sure what you mean by suffix too, if you mean the trailing `'/'` then that should be handled by setting the prefix to `list_objects_v2(..., Prefix='yourprefix/',...)`. I'll leave it up to you, just might be a bit cleaner.

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,135 @@
...
+                    "message": f"found message={message}",
+                },
+            )
+        return messages
+
+    def check_if_open(self) -> bool:
+        """
+        Check to see if the Post Office is open
+        """
+        if self.mesh_post_office_open == "True":
```
aparimoo-nhs, line 84: why not use `self.system_config["mesh_post_office_open"]` ?

terraform/modules/lambda/LR_25_mesh_post_office/iam.tf
jamesnunnnhs: Nope, terraform isn't my strong suit I trust you know what's best then.

terraform/modules/lambda/lambda.tf
```diff
@@ -8,6 +8,7 @@ locals {
     LR-12 = "LR_12_pds_registration_status"
     LR-15 = "LR_15_process_demo_diffs"
     LR-21 = "LR_21_split_dps_extract"
+    lr_25 = "LR_25_mesh_post_office"
```
aparimoo-nhs, line 11: not consistent with other names . see cases

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
joeds13: No you're right, I _was_ thinking if a file was put in on the key `yourprefix/dir/message.txt` it would pick up `yourprefix/dir/` as a key, but it'd only do that if someone/thing explicitly created that key, but that's fairly unlikely

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
joeds13: I followed the pattern of other uses of the the Spine common library by creating class variables of configuration and using them

terraform/modules/lambda/lambda.tf
joeds13: https://nhsd-confluence.digital.nhs.uk/display/PCRM/Terraform+Design+Choices

Inconsistent with existing, consistent with the design choices we all discussed and agree to try to adopt

It'd be a massive scope creep to start changing all of these on this PR to match, so I'm a fan of just writing new code as desired and uplifting old code as it's touched, makes sure we're not adding more debt that needs changing in the future

aparimoo-nhs: fair enough

test/unittests/lambdas/LR_25_mesh_post_office/conftest.py
```diff
@@ -0,0 +1,183 @@
...
+MOCK_INBOUND_KEY = "inbound_X00000"
+MOCK_OUTPUT_BUCKET = "lr-01-input"
+MOCK_OUTPUT_KEY = "inbound"
+REGION_NAME = os.environ.get("AWS_REGION")
+
+
+@pytest.fixture(autouse=True)
+def mesh_post_office_env_setup():
+    """
+    In ever test, set the environ to minimally mimic what Lambda runtime sets,
```
josh-scargill-nhsd, line 27: *every

test/unittests/lambdas/LR_25_mesh_post_office/test_LR_25_mesh_post_office.py
```diff
@@ -0,0 +1,73 @@
...
+    MOCK_INBOUND_KEY,
+    MOCK_OUTPUT_BUCKET,
+    MOCK_OUTPUT_KEY,
+)
+
+
+def test_check_if_open_should_be_open(mesh_post_office_open):
+    """
+    If the mesh_post_office_open parameter is truthy, we should expect the Post Office to be Open
+    """
```
josh-scargill-nhsd, line 15: I'm expecting the test name to be clear enough not to require a comment in most scenarios

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
joeds13: After finally getting a chance to look at this, setting `Prefix='yourprefix/'` still returns the object `yourprefix/`, so we'd still want to remove that from the returned items list, so going to leave this code as is

terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -9,6 +9,36 @@ locals {
...
+  mesh_kms_key_alias = {
+    default = ""
+    preprod = "alias/list-rec-preprod-mesh"
+    prod    = "alias/list-rec-prod-mesh"
+  }
+
+  mesh_post_office_open = {
+    default = "False"
+    preprod = "True"
+    preprod = "False"
```
aparimoo-nhs, line 21: should this be `prod`?

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,135 @@
...
+            return True
+        return False
```
jamesnunnnhs, line 86: Whenever I see string representations of true/false I see a bug. If it's possible to use a boolean value for this in terraform and it be easily passed then that would be good. If not, this is ok.

test/unittests/lambdas/LR_25_mesh_post_office/conftest.py
```diff
@@ -0,0 +1,183 @@
...
+        os.path.join(MOCK_INBOUND_KEY, DATA_FILE),
+    )
+    s3.upload_file(
+        os.path.join(DATA, DATA_FILE),
+        f"{MOCK_INBOUND_BUCKET}2",
+        os.path.join(f"{MOCK_INBOUND_KEY}2", DATA_FILE),
+    )
+
+
+@pytest.fixture(scope="class")
```
aparimoo-nhs, line 174: `class` ? i dont see any `class`? can it be `module` ?

test/unittests/lambdas/LR_25_mesh_post_office/conftest.py
```diff
@@ -0,0 +1,183 @@
...
+def s3():
+    with mock_s3():
+        yield boto3.client("s3", region_name=REGION_NAME)
+
+
+@pytest.fixture(scope="class")
```
aparimoo-nhs, line 180: may be `module`?

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
jamesnunnnhs: Righto

terraform/modules/lambda/lambda.tf
```diff
@@ -219,3 +220,17 @@ module "LR-24" {
   dynamodb_kms_key  = var.dynamodb_kms_key
   s3_kms_key        = var.s3_kms_key
 }
+
+module "lr_25" {
+  source = "./LR_25_mesh_post_office"
+
```
aparimoo-nhs, line 226: doesn't look it is formatted ? is the new line intentional

jamesnunnnhs (approved):
Approved with comment to implement if it's worth it/possible

terraform/modules/lambda/lambda.tf
joeds13: I find it's more to have a space between the modules meta variables (source/count/for_each etc) and it's variables

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
joeds13: Yeah fully agree, but in this case it's an SSM Parameter being picked up by the LambdaApplication class

SSM Params only have string types:
`String | StringList | SecureString`
https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_PutParameter.html#API_PutParameter_RequestSyntax

So there isn't away to treat it as a bool

test/unittests/lambdas/LR_25_mesh_post_office/conftest.py
joeds13: Set both to function (https://github.com/spulec/moto#example-on-pytest-usage)

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
```diff
@@ -0,0 +1,135 @@
...
+
+cwd: str = os.path.dirname(__file__)
+ADDITIONAL_LOG_FILE: str = os.path.join(cwd, "..", "..", "utils/cloudlogbase.cfg")
+
+
+class MeshPostOffice(LambdaApplication):
+    def __init__(self):
+        super().__init__(additional_log_config=ADDITIONAL_LOG_FILE, load_ssm_params=True)
+        self.mappings: dict = json.loads(str(self.system_config["mappings"]))
+        self.mesh_post_office_open: str = str(self.system_config["mesh_post_office_open"])
```
jamesnunnnhs, line 15: Environment vars are conventionally capitalised, which should be reflected here IMO

joeds13: These are not env vars, they are SSM Parameters and need to match directly, with full names for example:
- `/LR_25_mesh_post_office-pcrm-3538-bf63187e/mappings`
- `/LR_25_mesh_post_office-pcrm-3538-bf63187e/mesh_post_office_open`
I would not normally capitalise these

terraform/modules/lambda/variables.tf
```diff
@@ -1,3 +1,4 @@
+
```
josh-scargill-nhsd, line 1: 🚌

packages/lambda_code/LR_25_mesh_post_office/lr_25_lambda_handler.py
jamesnunnnhs: yep ok

terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -9,6 +9,36 @@ locals {
...
+          bucket = "list-rec-preprod-mesh"
+          key    = "inbound_X26OT179"
+        },
+        outbound = {
+          bucket = module.s3.buckets.LR-01.bucket
+          key    = "inbound"
+        }
+      }
+    ]
+    prod = []
```
josh-scargill-nhsd, line 39: Do we need to note anywhere that this will need populating at some point?

joeds13: Added to the Prod Env Setup ticket: https://nhsd-jira.digital.nhs.uk/browse/PCRM-3847

aparimoo-nhs (approved):

PR #50, opened 2021-07-15
PCRM-3517-Test-Coverage

premnhs:


test/integrationtests/step_impl/lr_15_lambda_cdd_file.py
```diff
@@ -55,23 +56,23 @@ def connect_to_s3_and_upload_mock_data_valid_scenario(lr_15_path):
...
-    with open(exp_cdd_path, "r") as exp_cdd_datafile:
-        exp_cdd_data = sorted(exp_cdd_datafile)
+    with open(exp_path, "r") as exp_datafile:
+        exp_data = sorted(exp_datafile)
 
         if result.get("Contents") != None:
             for key_items in result.get("Contents"):
                 for filename in key_items.items():
-                    if "-CDD-" in str(filename):
+                    if filetype in str(filename):
```
jamesnunnnhs, line 72: Why not use 
```
for filename in `result["Contents"].items():
    if filetype in filename:
        ...
```
You're already expecting that object to exist and the filename is always a string I think, and neither of those work then that's a test fail anyway isn't it?

test/integrationtests/step_impl/lr_23_s3_patient_payload_json_file.py
```diff
@@ -0,0 +1,69 @@
...
+    "connect to <lr_bucket> s3 bucket and ensure patient payload record file with patientid <patientid> is generated as expected <exp_datafile>"
+)
+def assert_expected_file_in_lr13(lr_bucket, patientid, exp_datafile):
+    s3 = dev.client("s3", REGION_NAME)
+    result = s3.list_objects(Bucket=LR_23_BUCKET)
+
+    if result.get("Contents") != None:
+        for key_items in result.get("Contents"):
+            for filename in key_items.items():
+                if patientid in str(filename):
```
jamesnunnnhs, line 50: same comment as above

test/integrationtests/step_impl/lr_15_lambda_cdd_file.py
premnhs: this was further fixed this way : 
`        for filename in result.get("Contents"):
            if patientid in filename:`

jamesnunnnhs (approved):

HarveyNHS (approved):
Approved

PR #51, opened 2021-07-15
PCRM-3988 - Add Point-In-Time-Recovery to DynamoDB

abbas-khan8:


terraform/modules/dynamodb/variable.tf
```diff
@@ -34,3 +34,8 @@ variable "attributes" {
 variable "table_range_key" {
   type = string
 }
+
+variable "recovery_flag" {
```
joeds13, line 38: I'd name this really obviously to `point_in_time_recovery_enabled` as it links directly to a named piece of AWS DynamoDB functionality, it'll make it obvious to future readers of this code

terraform/stacks/list-reconciliation/dynamodb.tf
```diff
@@ -20,6 +20,7 @@ module "jobs" {
       type = "S"
     }
   ]
+  recovery_flag = true
```
joeds13, line 23: I'd move this to line 6

I like to group the key + value and the key + object variables in terraform, it makes it easier and consistent to read and often looks better when terraform fmt'd

terraform/stacks/list-reconciliation/dynamodb.tf
```diff
@@ -34,6 +35,7 @@ module "jobs_stats" {
...
```
joeds13, line 38: I'd move this to line 32

I like to group the key + value and the key + object variables in terraform, it makes it easier and consistent to read and often looks better when terraform fmt'd

joeds13 (changes requested):

jamesnunnnhs (approved):
Nothing more to add beyond Joe's comments, lgtm

josh-scargill-nhsd:
You've tested this works in AWS right?

joeds13 (approved):

PR #54, opened 2021-07-19
PCRM-3980 Adding KMS Encryption to DynamoDB Tables

joeds13:
Adds a KMS module that can be invoked to create a key, alias and policy
for a use case


terraform/modules/kms/key.tf
```diff
@@ -0,0 +1,96 @@
...
+      "kms:Put*",
+      "kms:Update*",
+      "kms:Revoke*",
+      "kms:Disable*",
+      "kms:Get*",
+      "kms:Delete*",
+      "kms:TagResource",
+      "kms:UntagResource",
+      "kms:ScheduleKeyDeletion",
+      "kms:CancelKeyDeletion",
```
jamesnunnnhs, line 46: Since it's for admin is it not ok to do `kms:*`?

terraform/modules/kms/key.tf
```diff
@@ -0,0 +1,96 @@
...
+  statement {
+    sid       = "Allow use of the key"
+    effect    = "Allow"
+    resources = ["*"]
+
+    actions = [
+      "kms:Encrypt",
+      "kms:Decrypt",
+      "kms:ReEncrypt*",
+      "kms:GenerateDataKey*",
```
jamesnunnnhs, line 64: I think it's ok to wildcard an entire service, but I feel like using the wildcard is a bit gratuitous here, saving a couple of lines at the cost of being explicit.

terraform/modules/kms/key.tf
joeds13: These policies are taken verbatim from the key policy when creating a key via the Console, so I chose not to deviate from what AWS set (this same policy is also used within the Spine Mesh Module)

Also as we are only having this in a single place, and reusing the module, I think it's safe to be verbose here as it won't need to be repeated like we see in other places

terraform/modules/kms/key.tf
joeds13: See https://github.com/NHSDigital/list-reconciliation/pull/54#discussion_r673181526

jamesnunnnhs (approved):

josh-scargill-nhsd (approved):

PR #56, opened 2021-07-19
PCRM-3980 Adding KMS Encryption to S3 Buckets

joeds13:
Needs #54


jamesnunnnhs (commented):
Only the same comments as the other pr for KMS keys:

I think it's ok to wildcard an entire service, but I feel like using the wildcard is a bit gratuitous on e.g. "kms:ReEncrypt*", saving a couple of lines at the cost of being explicit.

joeds13:
> Only the same comments as the other pr for KMS keys:
> 
> I think it's ok to wildcard an entire service, but I feel like using the wildcard is a bit gratuitous on e.g. "kms:ReEncrypt*", saving a couple of lines at the cost of being explicit.

See #54 (comment)

jamesnunnnhs (approved):

josh-scargill-nhsd (approved):

PR #57, opened 2021-07-20
PCRM-0000 Integration Test Output Resource Helper

joeds13:
Return the resource name, in place of the value as when Gauge hits issues with missing mappings
it exits non-gracefully and gives no hint to the user as to which variable is missing, requiring
them to download the logs and find the error this way the tests should error as they won't be able
to lookup AWS resources on the name (with "MISSING_OUTPUT_" prefix)


test/integrationtests/step_impl/tf_aws_resources.py
```diff
@@ -9,3 +9,24 @@ def get_aws_resources():
...
         tf_output = json.load(json_data)
         return tf_output
+
+
+def get_terraform_outputs():
+    resources = {}
+    with open(OUTPUT_FILE) as json_data:
+        tf_output = json.load(json_data)
+        for key in tf_output:
+            resources[key] = tf_output[key]["value"]
```
jamesnunn, line 19: Ah go on:

```
{k: v["value"] for k, v in json.load(open(OUTPUT_FILE)).items()}
```

joeds13: 🐍

test/integrationtests/step_impl/tf_aws_resources.py
```diff
@@ -9,3 +9,19 @@ def get_aws_resources():
...
         return tf_output
+
+
+def get_terraform_outputs():
+    return {k: v["value"] for k, v in json.load(open(OUTPUT_FILE)).items()}
+
+
+def get_terraform_output(resource):
+    try:
+        return get_terraform_outputs()[resource]
```
jamesnunn, line 20: If you choose to implement my suggestion, you would probably want to just move that dict comprehension to here and be rid of the `get_terraform_outputs` function.

Also this should now just raise `KeyError` if the item doesn't exist - you can either let that happen or handle it with a nicer message

test/integrationtests/step_impl/tf_aws_resources.py
```diff
@@ -9,3 +9,16 @@ def get_aws_resources():
     with open(OUTPUT_FILE) as json_data:
         tf_output = json.load(json_data)
         return tf_output
+
+
+def get_terraform_output(resource):
+    try:
+        outputs = {k: v["value"] for k, v in json.load(open(OUTPUT_FILE)).items()}
+        return outputs[resource]
+    except:
```
jamesnunnnhs, line 18: Ok do `except KeyError`. If you use a bare exception and the json is somehow malformed, it will disguise the actual error. No need to go crazy, you could handle `IOError` and `JSONDecodeError` too but really the errors raised by those should be informative enough anyway.

jamesnunnnhs (approved):

premnhs (approved):
LGTM

aparimoo-nhs (approved):

PR #65, opened 2021-07-22
PCRM-3514 - Set logger's internal Id to Job Id

abbas-khan8:


jamesnunnnhs (approved):

packages/lambda_code/LR_15_process_demo_diffs/lr_15_lambda_handler.py
```diff
@@ -28,12 +28,13 @@ def __init__(self):
         self.job_id = None
 
     def initialise(self):
-        pass
+        self.job_id = self.event["job_id"]
```
aparimoo-nhs, line 31: you might want to add `try` `catch` block , spine library will not spit out error if  there is no `job_id`

packages/lambda_code/LR_12_pds_registration_status/lr_12_lambda_handler.py
```diff
@@ -27,6 +27,8 @@ def __init__(self):
     def initialise(self):
         self.job_id = self.event["job_id"]
 
+        self.log_object.set_internal_id(self.job_id)
```
aparimoo-nhs, line 30: you might want to add `try` `catch` block , spine library will not spit out error if  there is no `job_id`

packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
```diff
@@ -40,6 +40,8 @@ def __init__(self):
     def initialise(self):
         self.job_id = str(uuid4())
 
+        self.log_object.set_internal_id(self.job_id)
```
aparimoo-nhs, line 43: you might want to add `try` `catch` block , spine library will not spit out error if  there is no `job_id`

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -41,9 +42,13 @@ def start(self):
             .decode("utf-8")
         )
 
+        self.job_id = body["job_id"]
+
```
aparimoo-nhs, line 46: bit inconsistent,  some of the the `internal_ids` are getting `initialised` in `initialise` and some of them in start. I would suggest add them in `start` or  use `try` `catch` block

abbas-khan8: The one's initialised from start are the ones that are being passed the `job_id` directly in the lambda event, LR07 is slightly different as its triggered on an S3 event and extracts the id from the object body

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -21,9 +21,10 @@ def __init__(self):
         super().__init__(additional_log_config=ADDITIONAL_LOG_FILE)
         self.s3 = boto3.client("s3")
         self.lambda_ = boto3.client("lambda", region_name=self.system_config["AWS_REGION"])
+        self.job_id = None
 
     def initialise(self):
-        pass
+        self.log_object.set_internal_id(self._create_new_internal_id())
```
josh-scargill-nhsd, line 27: I don't this we should be doing this at all, we're going to just overwrite it asap

packages/lambda_code/LR_08_demographic_comparison/lr_08_lambda_handler.py
```diff
@@ -21,14 +21,30 @@ def __init__(self):
         self.patient_id = None
 
     def initialise(self):
-        self.job_id = str(self.event["job_id"])
-        self.patient_id = str(self.event["patient_id"])
+        self.log_object.set_internal_id(self._create_new_internal_id())
```
josh-scargill-nhsd, line 24: If you could do the job id here, why not do the job id here, then set the internal id below?

packages/lambda_code/LR_11_gp_registration_status/lr_11_lambda_handler.py
```diff
@@ -20,13 +20,29 @@ def __init__(self):
         self.job_id = None
 
     def initialise(self):
-        self.job_id = self.event["job_id"]
+        self.log_object.set_internal_id(self._create_new_internal_id())
```
josh-scargill-nhsd, line 23: same here

packages/lambda_code/LR_12_pds_registration_status/lr_12_lambda_handler.py
```diff
@@ -25,13 +25,29 @@ def __init__(self):
         self.job_id = None
 
     def initialise(self):
-        self.job_id = self.event["job_id"]
+        self.log_object.set_internal_id(self._create_new_internal_id())
```
josh-scargill-nhsd, line 28: same here

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
abbas-khan8: Some of the lambdas don't have access to `job_id` so just added them for consistency, but I've removed them from the ones that use `job_id`

HarveyNHS (approved):
Approved

josh-scargill-nhsd (approved):

PR #71, opened 2021-07-23
PCRM-0000 Remove S3 Security and Object Configurations

joeds13:
Remove object configs to see if this has any affect on the issues we're seeing with
destroying the environment
Remove the per bucket S3 Public Access Blocks after complaints of noise
in AWS Guard Duty from the security team, this can be replaced by
turning on the Public Access Blocks at the account level, which
automatically applies to all buckets


terraform/modules/s3/LR-01.tf
```diff
@@ -27,43 +27,26 @@ resource "aws_s3_bucket" "LR-01" {
...
-  block_public_policy     = true
-  ignore_public_acls      = true
-  restrict_public_buckets = true
-}
-
 resource "aws_s3_bucket_object" "inbound" {
-  bucket        = aws_s3_bucket.LR-01.id
-  key           = "inbound/"
-  acl           = "private"
-  force_destroy = true
```
jamesnunnnhs, line 43: Shouldn't force destroy be a var anyway?

joeds13: Yeah should have been `force_destroy = var.force_destroy` like on the buckets

joeds13: But we don't need it as the bucket is getting force destroyed, so wondering if this is a potential point of conflict during the terraform destroy

jamesnunnnhs (approved):

aparimoo-nhs (approved):

PR #75, opened 2021-07-26
PCRM-3979 Implement Lambda update

aparimoo-nhs:
Removed the missing additional logging file fix


jamesnunnnhs (commented):
Do we now need to supply any sort of config that defines any custom log message properties now we don't have cloudlogbase.cfg?

aparimoo-nhs:
> Do we now need to supply any sort of config that defines any custom log message properties now we don't have cloudlogbase.cfg?

We don't have our custom log file as yet. For now we using the default cloudlogbase.cfg. Could be done as a part of another story.

jamesnunnnhs (commented):
Does this update remove the need to define the `initialise` function too? If so, is it worth doing now?

josh-scargill-nhsd (approved):

joeds13:
Updating LR-25 Post Office Lambda not included in this PR

aparimoo-nhs:
> Does this update remove the need to define the `initialise` function too? If so, is it worth doing now?

No, there is no change in terms of `initialise`  also I think @abbas-khan8  is using `initialise ` for the logger now isn't?

HarveyNHS (approved):
Approved

jamesnunnnhs (approved):

PR #77, opened 2021-07-27
PCRM-0000 Add Account Level AWS Configs

joeds13:
Set a password policy, even though we have an SCP denying IAM user
creation, it stops false positive raising of this being insufficient
from Security Tests
Set the S3 Block Public Access controls at the account level, so we
don't need to set them individually on buckets


jamesnunnnhs (approved):

terraform/stacks/account/locals.tf
```diff
@@ -0,0 +1,15 @@
+locals {
+  name               = "list-reconciliation-${local.environment}"
+  environment        = lower(terraform.workspace)
+  data_clasification = local.environment == "prod" ? "5" : "1"
```
HarveyNHS, line 4: change clasification to classification

terraform/stacks/account/locals.tf
```diff
@@ -0,0 +1,15 @@
...
+
+  tags = {
+    TagVersion         = "1"
+    Programme          = "SpinePod5"
+    Project            = "ListReconciliation"
+    DataClassification = local.data_clasification
```
HarveyNHS, line 10: same here

terraform/stacks/account/terraform.tf
```diff
@@ -0,0 +1,29 @@
...
+}
+
+terraform {
+  backend "s3" {
+    bucket         = "terraform-list-reconciliation-state-mgmt"
+    key            = "account.tfstate"
+    dynamodb_table = "terraform-list-reconciliation-locks"
+    region         = "eu-west-2"
+    encrypt        = true
+    role_arn       = "arn:aws:iam::486319732046:role/LRTerraformBase"
```
HarveyNHS, line 27: Will this arn always be the same?

HarveyNHS (commented):
Please review comments

terraform/stacks/account/terraform.tf
joeds13: Yeah, this role is in our single management account, which stores our Terraform state and locks

HarveyNHS (approved):
Approved

PR #78, opened 2021-07-27
PCRM-3515 - Bug fix for reading file with no DOW records

abbas-khan8:


packages/gp_file_parser/parser.py
```diff
@@ -126,19 +132,17 @@ def _parse_row_pair(row_pair: RowPair) -> Row:
...
-        row_1[0] == RECORD_TYPE and row_1[1] == RECORD_1
-    ), f"Row 1 for all records must start with '{RECORD_TYPE}~{RECORD_1}'"
+    if row_1[0] != RECORD_TYPE or row_1[1] != RECORD_1:
+        raise InvalidGPExtract({INVALID_RECORD: INVALID_RECORD_TYPE})
 
-    assert (
-        row_2[0] == RECORD_TYPE and row_2[1] == RECORD_2
-    ), f"Row 2 for all records must start with '{RECORD_TYPE}~{RECORD_2}'"
+    if row_2[0] != RECORD_TYPE or row_2[1] != RECORD_2:
+        raise InvalidGPExtract({INVALID_RECORD: INVALID_RECORD_TYPE})
```
josh-scargill-nhsd, line 145: Is there a reason these aren't checked at the same time? it looks like the same output unless i'm losing it

packages/gp_file_parser/parser.py
```diff
@@ -198,12 +209,14 @@ def parse_gp_extract_text(
...
     Raises:
-        AssertionError: If the file is not a valid extract.
+        InvalidStructure: If the file is not a valid extract.
     """
 
     raw_text = [r.strip() for r in gp_extract_text.split("\n") if r]
 
-    assert raw_text[0] == r"503\*", r"Header must contain 503\*"
+    if raw_text[0] != r"503\*":
+        raise InvalidStructure(r"Header must contain 503\*")
```
josh-scargill-nhsd, line 218: contain or be exactly?

packages/gp_file_parser/validators.py
```diff
@@ -276,6 +284,41 @@ def surname(surname_val: str, **kwargs) -> ValidatedRecord:
...
+
+            non_punctuated_val = prev_surname_val.translate(
+                str.maketrans("", "", string.punctuation)
+            ).replace(" ", "")
+            if non_punctuated_val not in (None, ""):
+                if not non_punctuated_val.isupper():
+                    invalid_reason = INVALID_PREV_SURNAME
+
+        except (TypeError, ValueError):
+            invalid_reason = INVALID_PREV_SURNAME
```
josh-scargill-nhsd, line 317: If this is a dupe of the surname, can we reuse functionality?

packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
```diff
@@ -256,9 +257,9 @@ def handle_extract(self, prefix: InputFolderType, error_message: str = None):
         s3_client.delete_object(Bucket=bucket, Key=self.upload_key)
 
         if error_message:
-            timestamp = self.upload_date.strftime("%d%m%YT%H%M.%S.%f")
+            timestamp = self.upload_date.strftime("%Y%m%d%H%M%S")
```
josh-scargill-nhsd, line 260: just wondering why we've declared a new format for the file, is the other format vastly unreadable? or do we use this format later on down the line?

HarveyNHS (approved):
Approved

packages/gp_file_parser/validators.py
abbas-khan8: Yep I think we can, was going to pick this up with James at some point because it uses his validator which I'm not completely familiar with just yet and with refactor

packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
abbas-khan8: This was the format on the confluence page, after talks with Prem and Noor this is what was settled on. This is going to change now based on our conversation about job_id

packages/gp_file_parser/parser.py
abbas-khan8: You've lost it

josh-scargill-nhsd (approved):

PR #79, opened 2021-07-28
PCRM-0000 Adding Log Retention controls

joeds13:
Store logs in dev envs for short amount of time, go longer for
production


jamesnunnnhs (approved):

terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -3,6 +3,12 @@ locals {
   environment        = lower(terraform.workspace)
   data_clasification = local.environment == "prod" ? "5" : "1"
 
+  log_retention_in_days = {
+    default = 3
+    preprod = 14
+    prod    = 365
```
josh-scargill-nhsd, line 9: Does this need to be agreed or discussed anywhere?

joeds13: Confirming the Data Retention is listed as a Post MVP requirement

The Prod config here matches the defaults, so isn't really a change, just puts it in a now configurable location that should mean when the Post MVP decision on retention is made, it's a nice easy change here

josh-scargill-nhsd (approved):

abbas-khan8 (approved):

PR #86, opened 2021-07-30
PCRM-0000 Updating PDS FHIR API Tests

joeds13:
The API response from the Sandbox has now changed, it is responding with
"https://fhir.hl7.org.uk/CodeSystem/" as the URL for the value types,
rather than "https://fhir.nhs.uk/R4/CodeSystem/"


jamesnunnnhs (approved):

premnhs (approved):
LGTM

PR #96, opened 2021-08-02
PCRM-0000 Even shorter names

joeds13:
Further resources were attempting to be created with longer than the 64 character limit name


HarveyNHS (approved):
Approved

josh-scargill-nhsd (approved):

PR #97, opened 2021-08-02
PCRM-0000 Fix LR-15 Integration Test

joeds13:


HarveyNHS (changes requested):

test/integrationtests/step_impl/lr_15_lambda_cdd_file.py
```diff
@@ -1,94 +1,134 @@
...
-from .tf_aws_resources import get_terraform_output
-from utils.datetimezone import get_datetime_now
-
-import boto3
+import json
 import os
 import time
 
-# On github
-access_key = os.getenv("AWS_PUBLIC_KEY")
```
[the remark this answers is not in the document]
joeds13: I was basing it off the comment around this being used on Github and that `AWS_PUBLIC_KEY` and `AWS_PRIVATE_KEY` are non-standard Env variables for AWS `aws_access_key_id` and `aws_secret_access_key` respectively, so assumed it was all custom configuration in the previous Github builds

Are you able to checkout this PR locally and attempt running the tests on Windows please?

HarveyNHS (approved):
Approved

test/integrationtests/step_impl/lr_beforehooks.py
```diff
@@ -88,5 +81,47 @@ def before_hooks_lr23_bucket_data():
...
+    # Only retrieve the keys for each item in the table (minimize data transfer)
+    projectionExpression = ", ".join("#" + key for key in tableKeyNames)
+    expressionAttrNames = {"#" + key: key for key in tableKeyNames}
+
+    counter = 0
+    page = table.scan(
+        ProjectionExpression=projectionExpression, ExpressionAttributeNames=expressionAttrNames
+    )
+    with table.batch_writer() as batch:
+        while page["Count"] > 0:
```
jamesnunnnhs, line 114: This isn't explicitly decremented - is it updated when you call `page = table.scan(...`?

test/integrationtests/step_impl/test_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+        Messages.write_message(f"Object not found after waiting: {bucket}/{key}")
+        return False
+
+
+def get_latest_jobid():
+    job_table = dynamodb.Table(get_terraform_output("jobs_table"))
+    job_data = job_table.scan()
+    job_items = []
+    for key, value in job_data.items():
+        if key == "Items":
```
jamesnunnnhs, line 32: This will mask an implicit test (I presume) that `'Items'` must exist. Can you do this instead?

```job_items = [j for j in job_data['Items']]```

Doing that ^ will raise a `KeyError` if `'Items'` doesn't exist

File: test/integrationtests/step_impl/lr_beforehooks.py (hunk 2 above)
Comment (jamesnunnnhs) [reply test/integrationtests/step_impl/lr_beforehooks.py:114]: Looks like you  totally just stole that function from the top rated StackOverflow answer  so no prob, I trust them

File: test/integrationtests/step_impl/test_helpers.py [hunk 4]
```diff
@@ -0,0 +1,66 @@
...
+def create_timestamp():
+    now = get_datetime_now().now()
+    return "%02d%02d%02d%02d%d%d" % (
+        now.year,
+        now.month,
+        now.day,
+        now.minute,
+        now.second,
+        now.microsecond,
+    )
```
jamesnunnnhs, line 49: Best use `datetime.strftime`. Also I think you missed hour

return `now.strftime('%Y%m%d%H%M%S%f')`

test/integrationtests/step_impl/test_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+        status = describe_execution["status"]
+
+        if status == "SUCCEEDED":
+            return describe_execution
+        else:
+            poll_count += 1
+
+        if poll_count >= poll_limit:
+            Messages.write_message(f"LR 10 StepFunction arn: {execution_arn} status: {status}")
+            return describe_execution
```
jamesnunnnhs, line 64: Maybe raise an exception from this since it would be a fail for it not to succeed

test/integrationtests/step_impl/test_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+
+def await_s3_object_exists(bucket, key):
+    try:
+        waiter = s3.get_waiter("object_exists")
+        waiter.wait(Bucket=bucket, Key=key)
+        Messages.write_message(f"Object exists: {bucket}/{key}")
+        return True
+    except botocore.exceptions.WaiterError:
+        Messages.write_message(f"Object not found after waiting: {bucket}/{key}")
+        return False
```
jamesnunnnhs, line 24: Maybe raise an exception from this since it would be a fail for it not to succeed

test/integrationtests/step_impl/test_helpers.py
joeds13: Discussed with James to leave as is as it's consistent with the test code style

test/integrationtests/step_impl/test_helpers.py
joeds13: Discussed with James to leave as is as it's consistent with the test code style

jamesnunnnhs (approved):

PR #99, opened 2021-08-02
PCRM-3724 Add FHIR Call

aparimoo-nhs:
Integration with FHIR API PDS call 
Added service for SSM Get and Put parameters
Updated unit tests


test/unittests/packages/utils/test_ssm.py
```diff
@@ -0,0 +1,26 @@
...
+    param = get_ssm_params(SSM_PARAM_PREFIX, REGION_NAME)
+    expected = {
+        "pds_api_app_key": "test",
+        "pds_api_private_key": "'-----BEGIN PUBLIC KEY-----\n test\n-----END PUBLIC KEY-----'",
+        "pds_api_access_token": '{"access_token": "ZAbKUoARjkBAYFbfqdGdLtN9ucj6", "expires_in": "599", "token_type": "Bearer", "issued_at": "1627882540"}',
+    }
+    assert param == expected
+
+
+@pytest.mark.xfail
```
joeds13, line 18: Does this test need fixing?

test/unittests/packages/utils/test_ssm.py
```diff
@@ -0,0 +1,26 @@
...
+
+from test.unittests.conftest import SSM_PARAM_PREFIX, REGION_NAME
+
+
+def test_get_ssm(mock_pds_private_key, mock_pds_app_key, mock_pds_access_token):
+    param = get_ssm_params(SSM_PARAM_PREFIX, REGION_NAME)
+    expected = {
+        "pds_api_app_key": "test",
+        "pds_api_private_key": "'-----BEGIN PUBLIC KEY-----\n test\n-----END PUBLIC KEY-----'",
+        "pds_api_access_token": '{"access_token": "ZAbKUoARjkBAYFbfqdGdLtN9ucj6", "expires_in": "599", "token_type": "Bearer", "issued_at": "1627882540"}',
```
joeds13, line 13: Can the `access_token` look more like a test string than a real string?

As in `'{"access_token": "this_is_a_test_token"}'`

terraform/stacks/list-reconciliation/modules.tf
```diff
@@ -96,3 +108,9 @@ module "test_data" {
   LR_22_bucket = module.s3.buckets.LR-22.bucket
   kms_key_arn  = module.kms["s3"].output.arn
 }
+
+module "ssm_store" {
```
joeds13, line 112: Prefer to keep the module name to be the same as the modules source directory name, so just `ssm`

terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -15,6 +15,20 @@ locals {
     # prod = "arn:aws:iam::000000000000:role/LRTerraformDeploy"
   }
 
+  pds_fhir_api_uri = {
```
joeds13, line 18: These should be a single variable, for just the URL, as the paths won't change between environments spo can be stored as constants in Python

terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -15,6 +15,20 @@ locals {
...
+    dev     = "https://ref.api.service.nhs.uk/personal-demographics/FHIR/R4"
```
joeds13, line 19: We have no named environment `dev` so remove this

terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -15,6 +15,20 @@ locals {
...
+    default = "https://ref.api.service.nhs.uk/personal-demographics/FHIR/R4",
```
joeds13, line 20: Should default to the sandbox URL

terraform/modules/ssm/pds_ssm.tf
```diff
@@ -0,0 +1,23 @@
+resource "aws_ssm_parameter" "pds_fhir_api_private_key" {
```
joeds13, line 1: File can be called `pds.tf` as we are in the `ssm` module so don't need the repetition

terraform/modules/ssm/outputs.tf
```diff
@@ -0,0 +1,4 @@
+
+output "pds_ssm" {
```
joeds13, line 2: Should be more descriptive of the type, so something like `pds_ssm_parameters_path`

terraform/modules/lambda/variables.tf
```diff
@@ -72,3 +73,15 @@ variable "log_retention_in_days" {
     error_message = "The log_retention_in_days value must be greater than or equal to 1."
   }
 }
+
+variable "pds_token_url" {
```
joeds13, line 77: PLease include descriptions of these three variables

terraform/modules/lambda/lambda.tf
```diff
@@ -162,13 +165,16 @@ module "LR-12" {
...
   mock_pds_data_bucket_arn              = var.mock_pds_data_bucket.arn
-  pds_url                               = "s3://${var.mock_pds_data_bucket.name}/${var.pds_url}"
+  pds_url                               = var.pds_url
   pds_api_retries                       = var.pds_api_retries
   suffix                                = var.suffix
   lambda_handler                        = var.lambda_handler
   dynamodb_kms_key                      = var.dynamodb_kms_key
   s3_kms_key                            = var.s3_kms_key
   log_retention_in_days                 = var.log_retention_in_days
+  token_url                             = var.pds_token_url
```
joeds13, line 175: Keep the variable name consistent: `pds_token_url = var.pds_token_url`

terraform/modules/lambda/lambda.tf
```diff
@@ -69,12 +69,15 @@ module "LR-07" {
...
   lr_06_bucket             = var.s3_buckets.LR-06.bucket
   mock_pds_data_bucket_arn = var.mock_pds_data_bucket.arn
-  pds_url                  = "s3://${var.mock_pds_data_bucket.name}/${var.pds_url}"
+  pds_url                  = var.pds_url
   suffix                   = var.suffix
   lambda_handler           = var.lambda_handler
   dynamodb_kms_key         = var.dynamodb_kms_key
   s3_kms_key               = var.s3_kms_key
   log_retention_in_days    = var.log_retention_in_days
+  token_url                = var.pds_token_url
```
joeds13, line 78: Keep the variable name consistent: pds_token_url = var.pds_token_url

terraform/modules/lambda/lambda.tf
```diff
@@ -162,13 +165,16 @@ module "LR-12" {
...
+  pds_url                               = var.pds_url
   pds_api_retries                       = var.pds_api_retries
   suffix                                = var.suffix
   lambda_handler                        = var.lambda_handler
   dynamodb_kms_key                      = var.dynamodb_kms_key
   s3_kms_key                            = var.s3_kms_key
   log_retention_in_days                 = var.log_retention_in_days
+  token_url                             = var.pds_token_url
+  ssm_kms_key                           = var.ssm_kms_key
+  pds_ssm                               = var.pds_ssm_prefix
```
joeds13, line 177: Keep the variable name consistent: `pds_ssm_prefix = var.pds_ssm_prefix`

terraform/modules/lambda/lambda.tf
```diff
@@ -69,12 +69,15 @@ module "LR-07" {
...
-  pds_url                  = "s3://${var.mock_pds_data_bucket.name}/${var.pds_url}"
+  pds_url                  = var.pds_url
   suffix                   = var.suffix
   lambda_handler           = var.lambda_handler
   dynamodb_kms_key         = var.dynamodb_kms_key
   s3_kms_key               = var.s3_kms_key
   log_retention_in_days    = var.log_retention_in_days
+  token_url                = var.pds_token_url
+  ssm_kms_key              = var.ssm_kms_key
+  pds_ssm                  = var.pds_ssm_prefix
```
joeds13, line 80: Keep the variable name consistent: `pds_ssm_prefix = var.pds_ssm_prefix`

terraform/modules/lambda/LR-12/locals.tf
```diff
@@ -0,0 +1,3 @@
+locals {
+  param_name = "${var.lambda_name}-${var.suffix}"
```
joeds13, line 2: Variable should be called `name`

terraform/modules/lambda/LR-12/iam.tf
```diff
@@ -114,6 +115,18 @@ resource "aws_iam_policy" "policy" {
...
+        {
+            "Effect": "Allow",
+            "Action": [
+                "ssm:PutParameter",
+                "ssm:GetParametersByPath",
+                "ssm:GetParameters",
+                "ssm:GetParameter"
+            ],
+            "Resource":[
+                  "arn:aws:ssm:eu-west-2:${data.aws_caller_identity.current.account_id}:parameter/${var.pds_ssm}/**"
```
joeds13, line 128: As we have all three parameters created in terraform, we could have each named rather than using a wildcard

Then we could have separate statements, controlling fine grained controls on each of them, for example only allowing the put on the token variable

terraform/modules/lambda/LR-07/locals.tf
```diff
@@ -0,0 +1,3 @@
+locals {
+  ssm_name = "${var.lambda_name}-${var.suffix}"
```
joeds13, line 2: Just be `name`

terraform/modules/lambda/LR-07/iam.tf
```diff
@@ -114,8 +117,20 @@ resource "aws_iam_policy" "policy" {
...
+         {
+            "Effect": "Allow",
+            "Action": [
+                "ssm:PutParameter",
+                "ssm:GetParametersByPath",
+                "ssm:GetParameters",
+                "ssm:GetParameter"
+            ],
+            "Resource":[
+                  "arn:aws:ssm:eu-west-2:${data.aws_caller_identity.current.account_id}:parameter/${var.pds_ssm}/*"
```
joeds13, line 130: As we have all three parameters created in terraform, we could have each named rather than using a wildcard

Then we could have separate statements, controlling fine grained controls on each of them, for example only allowing the put on the token variable

pytest.ini
```diff
@@ -6,7 +6,9 @@ env =
     AWS_SECRET_ACCESS_KEY=testing
     AWS_SECURITY_TOKEN=testing
     AWS_SESSION_TOKEN=testing
-    PDS_API_URL=s3://mock-pds-data/pds_api_data.csv
+    PDS_API_URL=http://test-url
+    PDS_TOKEN_URL=https://test-url/oauth/token
+    SSM_STORE_PREFIX=/ash-dev/pds/
```
joeds13, line 11: Generic name not `ash-dev`

packages/utils/ssm.py
```diff
@@ -0,0 +1,28 @@
+import boto3
+import os
+import json
+
+
+def get_ssm_params(ssm_path, region):
+    ssm_client = boto3.client("ssm", region_name=region)
```
joeds13, line 7: Initialise the ssm client in the top level scope and then use it in the two functions

Means boto3 is initialised at application startup rather than every function call

packages/utils/pds_api_service.py
```diff
@@ -1,120 +1,380 @@
...
-        "version": content["meta"]["versionId"],
-    }
-
-
-def get_pds_record(nhs_number, *args, **kwargs):
-    if PDS_API_URL.startswith("http"):
-        return get_pds_fhir_record(PDS_API_URL, nhs_number, *args, **kwargs)
-    elif PDS_API_URL.startswith("s3"):
-        return get_mock_pds_record(PDS_API_URL, nhs_number, *args, **kwargs)
+            # TODO Change to prod before merge
```
joeds13, line 242: Do this TODO?

packages/utils/pds_api_service.py
```diff
@@ -1,120 +1,380 @@
...
+        {"name": "sensitive", "value": "meta.security.[*].code", "default": ""},
+        {"name": "version", "value": "meta.versionId", "default": ""},
+    ]
+}
+
+
+class FHIRAPIHelper:
+    """Class for FHIR API request"""
+
+    def __init__(self, system_config, logger):
```
joeds13, line 74: Passing in the LambdaApplication system_config and logger tightly couples this code to the spine-aws-common implementation

It's a significant change but it we be better to only pass in the configuration specific to the FHIR API

Also it's debatable if there should be any logging in this class, as it's not a "business logic" piece of code and should just be passing errors back up to the caller

packages/utils/pds_api_service.py
```diff
@@ -1,120 +1,380 @@
...
+            logger.write_log(
+                "UTI9995",
+                "Lambda system config unavailable",
+                {
+                    "logger": "List-Rec FHIR Service",
+                    "level": "INFO",
+                    "message": "System config is not available",
+                },
+            )
+            raise PDSAPIError("System config not available, exiting")
```
joeds13, line 86: There's some naming inconsistency that may confuse users inexperienced with this area with having both `PDSAPI` and `FHIRAPI` used interchangeably

joeds13 (changes requested):

packages/utils/pds_api_service.py
```diff
@@ -1,120 +1,380 @@
...
+                    headers=additional_headers,
+                )
+                form_data = {
+                    "client_assertion_type": "urn:ietf:params:oauth:client-assertion-type:jwt-bearer",
+                    "client_assertion": j,
+                    "grant_type": "client_credentials",
+                }
+                response = requests.post(self.pds_token_url, data=form_data)
+                token = response.json()
+                put_ssm_params(
```
joeds13, line 134: Are we worried about any race conditions here, if this called in two Lambdas, will they fight? Would a collision even cause an issue?

terraform/modules/lambda/variables.tf
```diff
@@ -48,6 +48,7 @@ variable "s3_kms_key" {
   type = map(string)
 }
 
+
```
abbas-khan8, line 51: Unnecessary extra line here?

HarveyNHS (approved):
Approved

packages/utils/pds_api_service.py
aparimoo-nhs: I didn't see any issue so far , looking at forum some have noticed exception 'TooManyUpdate'. i have handled that in code to retry if '`TooManyUpdate`' exception is raised

terraform/stacks/list-reconciliation/locals.tf
aparimoo-nhs: that would tie code with url if the version changes , the way it looks that version no is appended to the path. if version changes code needs changing

joeds13: If the version in the API URI changes, then that will be a breaking contract change which would result in the code needing to change? 
Therefore it makes sense that the code owns what version(s) it supports rather than it just being passed along in the URL as part of the configuration

terraform/modules/ssm/pds.tf
```diff
@@ -0,0 +1,23 @@
+resource "aws_ssm_parameter" "pds_fhir_api_private_key" {
+  name      = "/${var.prefix}/pds/pds_api_private_key"
+  overwrite = true
+  type      = "SecureString"
+  value     = "test"
```
joeds13, line 5: `value     = "To Replace"`

Matches MESH Module: https://github.com/NHSDigital/spine-core-aws-common/blob/develop/terraform/mesh_aws/ssm_config.tf

terraform/modules/ssm/pds.tf
```diff
@@ -0,0 +1,23 @@
...
+  key_id    = var.ssm_kms_arn
+}
+
+resource "aws_ssm_parameter" "pds_fhir_app_key" {
+  name      = "/${var.prefix}/pds/pds_api_app_key"
+  overwrite = true
+  type      = "SecureString"
+  value     = "test"
```
joeds13, line 21: `value     = "To Replace"`

Matches MESH Module: https://github.com/NHSDigital/spine-core-aws-common/blob/develop/terraform/mesh_aws/ssm_config.tf

abbas-khan8 (approved):

terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -15,6 +15,13 @@ locals {
     # prod = "arn:aws:iam::000000000000:role/LRTerraformDeploy"
   }
 
+  pds_fhir_api_uri = {
+    test    = "https://ref.api.service.nhs.uk"
+    default = "https://sandbox.api.service.nhs.uk",
+    preprod = ""
```
joeds13, line 21: If this is empty, Integration Tests in PreProd will fail, should it be sandbox or ref?

If it's ref we'll need to get the creds setup after merge

As you're editing it, I'd call this variable `pds_fhir_api_url` not `uri` to be more accurate

joeds13 (changes requested):

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -118,11 +119,11 @@ def pds_hydrate(self, job_id: str, record: Demographics) -> Success:
...
 
         try:
-            pds_record = get_pds_record(record.NhsNumber, max_retries=5, backoff_factor=1)
+            pds_record = self.api.get_pds_record(record.NhsNumber, job_id)
 
         except PDSAPIError as err:
-            msg = f"Error fetching PDS record for NHS number {record.NhsNumber}, {err.details}"
-            error_response = log_dynamodb_error(self.log_object, job_id, err.details["code"], msg)
+            msg = f"Error fetching PDS record for NHS number {record.NhsNumber}, {str(err)}"
+            error_response = log_dynamodb_error(self.log_object, job_id, str(err), msg)
```
jamesnunnnhs, line 126: Are we still supposed to be logging errors to dynamo? I thought we were dropping that

packages/utils/pds_api_service.py
```diff
@@ -1,120 +1,314 @@
...
+    ]
+}
+
+
+class PDSAPIHelper:
+    """Class for FHIR PDS API request"""
+
+    def __init__(self, system_config):
+        self.pds_data = {}
+        if not isinstance(system_config, dict):
```
jamesnunnnhs, line 82: How about adding a typehint for system_config and a bit of documentation about how it's used or behaves.

packages/utils/pds_api_service.py
```diff
@@ -1,120 +1,314 @@
...
+                        string_type="SecureString",
+                    )
+                except RuntimeError:
+                    # Avoid Race condition , try again to add parameters
+                    put_ssm_params(
+                        f"{self.ssm_store_path}{PDSParamStore.PDS_ACCESS_TOKEN.value}",
+                        region="eu-west-2",
+                        data_string=token,
+                        string_type="SecureString",
+                    )
```
jamesnunnnhs, line 137: Isn't it possible for this to enter a non-exiting loop until the lambda times out? Maybe put a retry on this? 

See `utils.retry_func` for a retryer

packages/utils/pds_api_service.py
```diff
@@ -1,120 +1,314 @@
...
+            else:
+                # token is either expired or not set
+                token = json.loads(self.ssm_params["pds_api_access_token"])
+
+            return token
+
+        except Exception as e:
+            raise PDSAPIError(e)
+
+    def get_pds_record(self, nhs_number, job_id) -> dict:
```
jamesnunnnhs, line 147: Can you add some typehints for this and if it's possible, a return type as expected like `Dict[str, str]` or even define a type and return that?

packages/utils/pds_api_service.py
```diff
@@ -1,120 +1,314 @@
...
+            # Patient Record is REDACTED
+            redacted_patient_data = self.convert_pds_to_list_rec_data()
+            redacted_patient_data.update(
+                {"sensitive": SensitiveMarkers.REDACTED.value}
+            )  # update dict with REDACTED
+            return redacted_patient_data
+
+        elif status_code == 404 and error_code == "RESOURCE_NOT_FOUND":
+            # PDS data not found for the patient
+            return {}
```
jamesnunnnhs, line 285: I would tend to return `None` rather than an empty object unless you'd expect to be able to process the empty object later in the caller

pytest.ini
```diff
@@ -6,7 +6,8 @@ env =
     AWS_SECRET_ACCESS_KEY=testing
     AWS_SECURITY_TOKEN=testing
     AWS_SESSION_TOKEN=testing
-    PDS_API_URL=s3://mock-pds-data/pds_api_data.csv
+    PDS_BASE_URL=http://sandbox.test-url
+    SSM_STORE_PREFIX=/dev/pds/
```
jamesnunnnhs, line 10: should the prefix include the trailing `/`? I think so because it looks like in the iam definition you add it explicitly

terraform/modules/lambda/LR-12/iam.tf
```diff
@@ -114,6 +115,26 @@ resource "aws_iam_policy" "policy" {
...
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "ssm:GetParametersByPath",
+                "ssm:GetParameters",
+                "ssm:GetParameter"
+            ],
+            "Resource":[
+                  "arn:aws:ssm:eu-west-2:${data.aws_caller_identity.current.account_id}:parameter/${var.pds_ssm_prefix}/**"
```
jamesnunnnhs, line 127: Two `*`s? Does it handle that or is it an error?

terraform/modules/lambda/LR-12/lambda.tf
```diff
@@ -32,8 +33,9 @@ resource "aws_lambda_function" "LR-12-Lambda" {
       ERRORS_TABLE                            = var.errors_table_name
       LR_13_REGISTRATIONS_OUTPUT_BUCKET       = var.registrations_output_bucket
       LR_22_PDS_PRACTICE_REGISTRATIONS_BUCKET = var.pds_practice_registrations_bucket
-      PDS_API_URL                             = var.pds_url
+      PDS_BASE_URL                            = var.pds_base_url
       PDS_API_RETRIES                         = var.pds_api_retries
+      SSM_STORE_PREFIX                        = "/${var.pds_ssm_prefix}/"
```
jamesnunnnhs, line 38: Just check the prefix trailing `/` per my other comments

terraform/modules/lambda/lambda.tf
```diff
@@ -70,12 +70,16 @@ module "LR-07" {
   lr_06_bucket_arn         = var.s3_buckets.LR-06.arn
   lr_06_bucket             = var.s3_buckets.LR-06.bucket
   mock_pds_data_bucket_arn = var.mock_pds_data_bucket.arn
-  pds_url                  = "s3://${var.mock_pds_data_bucket.name}/${var.pds_url}"
   suffix                   = var.suffix
   lambda_handler           = var.lambda_handler
   dynamodb_kms_key         = var.dynamodb_kms_key
   s3_kms_key               = var.s3_kms_key
   log_retention_in_days    = var.log_retention_in_days
+
```
jamesnunnnhs, line 78: I think the terraform gods won't like this newline

terraform/modules/ssm/outputs.tf
```diff
@@ -0,0 +1,16 @@
+
+output "pds_ssm_parameters_path" {
+  value = "${var.prefix}/pds"
```
jamesnunnnhs, line 3: Check how this plays with the prefix having included the trailing `/`

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
aparimoo-nhs: not in this story

pytest.ini
aparimoo-nhs: This is for test so i am passing the prefix

joeds13 (approved):

PR #104, opened 2021-08-03
PCRM-3972 Run LR-09 on a Schedule

joeds13:
Allow passing in of rate expressions and conditionally create schedules
when desired
Also uplift LR-25 to follow the same configurable pattern

For the demo, setting the schedule to every minute (AWS doesn't allow you to set it lower) by:

`terraform/stacks/list-reconciliation/locals.tf`

```
lr_09_event_schedule_expression = {
    default = "rate(1 minute)"
    preprod = "rate(5 minutes)"
    prod    = "rate(5 minutes)"
  }
```


josh-scargill-nhsd:
Just an FYI, this should be off by default

josh-scargill-nhsd (approved):

abbas-khan8 (approved):

PR #108, opened 2021-08-04
Fix branch id command

joeds13:
Move branch name trimming into the command, currently it's not being
picked up in the env destroy workflow, due to how those workflows have
to pickup the branch name, meaning they're not destroying successfully
as the name is not matching the existing environment


HarveyNHS (approved):
Approved

jamesnunnnhs (approved):

PR #109, opened 2021-08-04
PCRM-3514 - Refactored lambdas using new logging method

abbas-khan8:


packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
```diff
@@ -1,88 +1,61 @@
...
 
+cwd = os.path.dirname(__file__)
+ADDITIONAL_LOG_FILE = os.path.join(cwd, "..", "..", "utils/cloudlogbase.cfg")
+
 
 class ValidateAndParse(LambdaApplication):
     def __init__(self):
-        super().__init__()
+        super().__init__(additional_log_config=ADDITIONAL_LOG_FILE)
+        self.s3 = boto3.client("s3")
```
josh-scargill-nhsd, line 28: Just curious, why have we done this? Have I missed a recommendation that we should be setting boto3 clients in the init?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -33,144 +40,74 @@ def start(self):
...
             prefix = f"{InputFolderType.FAIL.value}logs/"
 
-            self.log_object.set_internal_id(self._create_new_internal_id())
             self.log_key = self.event["Records"][0]["s3"]["object"]["key"]
             self.log_filename = str(self.log_key).replace(prefix, "")
 
-            self.response = self.process_failed_file()
+            try:
+                self.job_id = self.log_filename.split("-FailedFile-")[1].replace(".json", "")
+                UUID(str(self.job_id))
```
josh-scargill-nhsd, line 48: Why are we doing this?

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -33,144 +40,74 @@ def start(self):
...
-            self.log_object.write_log(
-                "UTI9998",
-                sys.exc_info(),
-                {
-                    "logger": "LR04.Lambda",
-                    "level": "ERROR",
-                    "message": error_message,
-                },
+            self.response = error(
+                f"LR04 Lambda tried to access missing key={str(err)}", self.log_object.internal_id
```
josh-scargill-nhsd, line 73: Are these thrown as critical errors? I was under the impression responses weren't logged errors

abbas-khan8: No, nothing is thrown or logged, this is just the response of the lambda. We wouldn't log in this case since the base Lambda Application catches and logs all exceptions. But it's probably worth changing the response so we know exactly which exception caused the lambda to stop rather than the default "Lambda application has stopped" response

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
abbas-khan8: The code extracts the job id from the filename and attempts to convert it into a UUID. It throws a value error if the string given is not a valid UUID

packages/lambda_code/LR_02_validate_and_parse/lr_02_lambda_handler.py
abbas-khan8: Rather than repeating calls to boto3 for resources, if the same resource is used in multiple places it's better to initialise it once in the init and use it where necessary

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
josh-scargill-nhsd: Is there a clearer way of demonstrating that's what you're trying to achieve?

abbas-khan8: Other ways are a long regex to test for a UUID match but it would look pretty ugly and that's kind of what `UUID(str)` does. Otherwise a comment to explain it?

packages/utils/cloudlogbase.cfg
```diff
@@ -0,0 +1,327 @@
...
+Log Level = INFO
+Log Text = uploadFile='{upload_filename}' has failed validation for jobId={job_id}
+
+[LR02C01]
+Log Level = CRITICAL
+Log Text = Unable to write details to the database for jobId='{job_id}'
+
+[LR02C02]
+Log Level = CRITICAL
+Log Text = records='{num_of_records}' failed to process for jobId={job_id}'
```
josh-scargill-nhsd, line 55: missing a `'`

packages/utils/cloudlogbase.cfg
```diff
@@ -0,0 +1,327 @@
...
+Log Level = INFO
+Log Text = Gathering output files for jobId='{job_id}'
+
+[LR14I02]
+Log Level = INFO
+Log Text = Successfully sent output files='{count}' to meshId='{mesh_id}' for workflowId='{workflow_id}' for jobId='{job_id}'
+
+[LR14I03]
+Log Level = INFO
+Log Text = Notified emailAddress='{receiving_address}' with subject='{subject} for jobId='{job_id}'
```
josh-scargill-nhsd, line 175: and here too

packages/utils/cloudlogbase.cfg
```diff
@@ -0,0 +1,327 @@
...
+
+[LR15I01]
+Log Level = INFO
+Log Text = Processing patients='{count}' with demographic differences for jobId='{job_id}'
+
+[LR15I02]
+Log Level = INFO
+Log Text = exported={count} records sent to the DSA for jobId={job_id}
```
josh-scargill-nhsd, line 183: I'm confused with the consistency, should we be putting `'` around `{vars}` or not?

q
```diff
@@ -0,0 +1,141 @@
+* [31mce4f7cb[m[33m (HEAD -> PCRM-3514-logging-refactor, origin/PCRM-3514-logging-refactor)[m updated logs[32m(8 minutes ago)[1;34m<abbas-khan8>[m
+* [31mef84231[m[33m[m PCRM-3514 - Refactor lambda loggers[32m(8 hours ago)[1;34m<abbas-khan8>[m
+* [31mea27e96[m[33m[m PCRM-4079 Adding KMS Encryption to Cloudwatch Logs[32m(10 hours ago)[1;34m<Joe Stringer>[m
+* [31ma20f546[m[33m[m PCRM-4079 Disable TFSec Lambda XRay check[32m(10 hours ago)[1;34m<Joe Stringer>[m
```
josh-scargill-nhsd, line 4: what's this?

terraform/modules/firehose/iam.tf
```diff
@@ -0,0 +1,65 @@
...
+      "Action": "sts:AssumeRole",
+      "Principal": {
+        "Service": "logs.${data.aws_region}.amazonaws.com"
+      },
+      "Effect": "Allow"
+    }
+  ]
+}
+EOF
+
```
josh-scargill-nhsd, line 22: I thought your firehose work was coming as a separate ticket?

abbas-khan8: It will be, this has snuck in from another branch

q
abbas-khan8: I'm not sure why this appears, its an auto-generated file that finds its way into my listrec folder when I run tests

packages/utils/cloudlogbase.cfg
abbas-khan8: Yep we should, they were missing from these ones. After manually adding them you miss a few every now and then

packages/lambda_code/LR_04_feedback_failure/lr_04_lambda_handler.py
josh-scargill-nhsd: Maybe a comm to explain as it is out of context here

jamesnunnnhs (approved):

packages/lambda_code/LR_14_send_list_rec_results/lr_14_lambda_handler.py
```diff
@@ -11,44 +11,83 @@
 from spine_aws_common.lambda_application import LambdaApplication
 from utils.database.models import DemographicsDifferences, Jobs, JobStats
 from utils.datetimezone import localize_date
-from utils.logger import success
 from utils.ssm import get_ssm_params
+from utils.logger import success, error, Message
+
+
+
```
josh-scargill-nhsd: :bus:

josh-scargill-nhsd (approved):

PR #110, opened 2021-08-05
PCRM-4079 Pen Test Remediation

joeds13:
Resolving issues raised by TFSec


jamesnunnnhs (approved):

terraform/modules/kms/key.tf
```diff
@@ -93,4 +93,37 @@ data "aws_iam_policy_document" "key" {
       identifiers = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/NHSDAdminRole"]
     }
   }
+
+  dynamic "statement" {
+    # if this module is for the cloudwatch KMS key
+    for_each = length(regexall(".*cloudwatch.*", var.name)) > 0 ? [var.name] : []
+
+    content {
+      sid       = "CloudwatchAllow"
```
HarveyNHS, line 102: Does this need to be CloudWatchAllow with a capital W?

HarveyNHS (commented):
Please review

terraform/modules/kms/key.tf
joeds13: It doesn't need it, `sid` is just a friendly name

Also as an FYI, Actions in IAM policies are not case sensitive anyway

aparimoo-nhs (approved):

PR #112, opened 2021-08-05
PCRM-0000 - Remove postcode character limit

abbas-khan8:


joeds13 (approved):

jamesnunnnhs:
Let's just keep this on ice until there's confirmation that we will loosen the validation rules then do them all at once - turns out there's more to do like handle sex as m/f and allow shorter ha ciphers. There may be more but that's what was found

PR #114, opened 2021-08-06
PCRM-3538-Test Coverage for preprod

HarveyNHS:
Ready for review


joeds13:
> Currently the last test for "failed" doesn't work at the moment - I don't know how to call the filename correct when it looks into the failed/logs folder - If you can figure it out and place it up for PR by all means
> 
> I also tried to configure the tests to run on preprod but the terraform seemed to have failed in the process, if you can help fix that would be great.

I've pushed a commit, that sorts out the running of the test on PreProd

I also don't think we need to do the "failed" test, LR-25 the code under test, is only concerned with moving any file it finds in it's configured source bucket to it's configured target bucket, it will be doing this for both GP data and DPS PDS data. 
Therefore it's out of scope for LR-25 to care if a downstream service deems the file it has moved to process successfully or not. This overall functionality will be within scope of the end to end test(s).

packages/utils/__init__.py
```diff
@@ -16,6 +16,8 @@ class InputFolderType(Enum):
     PASS = "pass/"
     FAIL = "fail/"
     RETRY = "retry/"
+    MESH_INBOUND = "inbound_X26OT179/"
+    MESH_OUTBOUND = "outbound_X26OT178_to_X26OT179/"
```
jamesnunnnhs, line 20: @joeds13 shouldn't this be evaluated during TF build? We have the individual values in `list-reconciliation/locals.tf locals->mesh_mappings` already so could construct and output these?

test/integrationtests/step_impl/lr_25_preprod_mesh_inbound.py
```diff
@@ -0,0 +1,69 @@
...
+        Messages.write_message("File not found")
+        raise
+
+
+@step("invoke lambda LR-25")
+def invoke_lambda_lr_25():
+    response = lambda_.invoke(FunctionName=LR_25_LAMBDA_ARN, LogType="Tail", Payload=json.dumps({}))
+    for key, value in response.items():
+        if key == "ResponseMetadata":
+            assert value["HTTPStatusCode"] == int(200)
```
jamesnunnnhs, line 52: Might be worth `int`ing the status code too to make sure you're getting what you expect.

packages/utils/__init__.py
joeds13: Yeah, definitely need to be pulled from Terraform as the MESH mailboxes are environment specific

josh-scargill-nhsd (dismissed):

joeds13 (dismissed):
Need to pull MESH values from Terraform Outputs

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -83,13 +83,14 @@ def start(self):
...
 
-        except PDSAPIError:
+        except PDSAPIError as err:
             self.log_object.write_log(
                 "LR07C01",
                 log_row_dict={
                     "patient_id": patient.Id,
                     "nhs_number": patient.NhsNumber,
                     "job_id": self.job_id,
+                    "response_message": str(err),
```
jamesnunnnhs, line 93: Could you make these `traceback.format_exc()`? In the absence of errors being raise, this will show the entire traceback instead of just the message

packages/lambda_code/LR_12_pds_registration_status/lr_12_lambda_handler.py
```diff
@@ -136,6 +136,7 @@ def get_pds_exclusive_registrations(self) -> Message:
                         log_row_dict={
                             "nhs_number": nhs_number,
                             "job_id": self.job_id,
+                            "response_message": str(err),
```
jamesnunnnhs, line 139: Could you make these `traceback.format_exc()`? In the absence of errors being raise, this will show the entire traceback instead of just the message. (I know I made the pr that you included for this)

jamesnunnnhs (approved):
Approving provided the error message being logged is updated

jamesnunnnhs (approved):

PR #115, opened 2021-08-09
PCRM-3973 Incomplete Reconciliation Check

joeds13:
Jobs are marked as timed out and removed from In Flight after a
configurable amount of time


packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
```diff
@@ -1,27 +1,27 @@
...
 from utils.database.models import Jobs, InFlight, Demographics, JobStats
-from utils.logger import success, log_dynamodb_error
+from utils.logger import log_dynamodb_error
 from utils.statuses import JobStatus
 
 
 class ScheduledCheck(LambdaApplication):
     def __init__(self):
         super().__init__()
+        self.job_timeout_hours = int(str(self.system_config["JOB_TIMEOUT_HOURS"]))
```
jamesnunnnhs, line 16: why not just `int(...` instead of `int(str(...`?

packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
```diff
@@ -40,67 +40,78 @@ def start(self):
...
         response = Demographics.JobIdIndex.count(
             job_id, filter_condition=Demographics.IsComparisonCompleted == True
         )
-
         return response == total_records
 
+    @staticmethod
+    def is_job_timed_out(timestamp: datetime, job_timeout_hours: int) -> bool:
+        cutoff_time = datetime.now() - timedelta(hours=job_timeout_hours)
+        return timestamp.replace(tzinfo=None) < cutoff_time
```
jamesnunnnhs, line 55: Need to be careful with `datetime.now()`s because of the unapplied (or was it applied) daylight savings time in the lambda container plus the actual timezone set. Can't remember the details but Abbas solved it with the function at `utils.datetimezone.get_datetime_now`. Unittests may well work but then on AWS it'll break somehow.

packages/utils/database/models.py
```diff
@@ -179,4 +182,5 @@ class Meta:
         write_capacity_units = 1
 
     JobId = UnicodeAttribute(hash_key=True)
+    Timestamp = UTCDateTimeAttribute(default=datetime.utcnow())
```
jamesnunnnhs, line 185: I think this needs to be `GMTDateTimeAttribute`

test/unittests/lambdas/LR_09_scheduled_check/test_scheduled_check.py
```diff
@@ -48,29 +44,50 @@ def test_three_inflight_records_one_processed_correctly_two_skipped(
...
+def test_is_job_timed_out_timestamp_less_than_job_timeout_hours(lambda_handler):
+    timestamp = datetime.now()
+    job_timeout_hours = 1
+    assert lambda_handler.is_job_timed_out(timestamp, job_timeout_hours) == False
+
+
+def test_is_job_timed_out_timestamp_more_than_job_timeout_hours(lambda_handler):
+    timestamp = datetime.now() - timedelta(hours=2)
+    job_timeout_hours = 1
+    assert lambda_handler.is_job_timed_out(timestamp, job_timeout_hours) == True
```
jamesnunnnhs, line 87: Not really necessary to change but for brevity you can just `assert lambda_handler.is_job_timed_out(timestamp, job_timeout_hours)` because the result of that function is already boolean

packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
joeds13: When just `int(self.system_config["JOB_TIMEOUT_HOURS"])`

We get from the linter:
```
No overloads for "__new__" match the provided arguments
  Argument types: (str | dict[Unknown, Unknown] | bytes)PylancereportGeneralTypeIssues
```

So it gets upset wanting to convert straight to `int` even though we know it'd be okay in this case

So mainly doing both to silence the linter, but it kind of makes sense as we know it's an env var which is string like, so we say it's definitely a string then we want it to be an int

packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
joeds13: `utils.datetimezone.get_datetime_now` returns times in "Europe/London" timezone, we should be using UTC throughout the code, only converting to local time zones if needed when projecting a date/time, such as in an email to an end user.

AWS Lambdas run as UTC and this is consistent with the use of UTC times at the creation of the Job timestamp

packages/utils/database/models.py
joeds13: See https://github.com/NHSDigital/list-reconciliation/pull/115#discussion_r685780543

test/unittests/lambdas/LR_09_scheduled_check/test_scheduled_check.py
joeds13: I'd agree in non-test code

But here I like that you can look down the set of functions and easily see the `False/True/False` on each assertion for the different test cases

packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
jamesnunnnhs: That is bonkers but ok

packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
jamesnunnnhs: Well in that case we have a situation where we have differing timezone elements in the code - not sure what the impact really is but @josh-scargill-nhsd worth worrying about?

josh-scargill-nhsd: As of right now the entire codebase uses BST, if we want to store in UTC and convert on retrieval (sensible) then we need to refactor everywhere we do it.
So basically either confirm to how it's currently done, or refactor it all.

joeds13: See https://github.com/NHSDigital/list-reconciliation/pull/115/commits/f846d4a86cae0dd625439b8c5486a55444838351 for commit refactoring to UTC

HarveyNHS (approved):
Approved

joeds13:
It should be expected that systems run in UTC, it's the same as half our year anyway, and negates the need for a lot of logic to handle daylight saving times and the potential bugs of the system processing data during the extra hour or hour skip.

> * LR-04

Added Timezone "(UTC)" to the email output string

> * LR-14

Added Timezone "(UTC)" to the email output string

> * Logging? or is that fine in UTC?

UTC preferred

packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
```diff
@@ -40,67 +40,78 @@ def start(self):
...
     def is_job_complete(job_id: str, total_records: int) -> bool:
         response = Demographics.JobIdIndex.count(
             job_id, filter_condition=Demographics.IsComparisonCompleted == True
         )
-
         return response == total_records
 
+    @staticmethod
+    def is_job_timed_out(timestamp: datetime, job_timeout_hours: int) -> bool:
+        cutoff_time = datetime.now() - timedelta(hours=job_timeout_hours)
```
josh-scargill-nhsd, line 54: This `datetime.now()` should use the default `get_datetime_now()` since it's aligned with what you're trying to do now anyway

packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
```diff
@@ -40,67 +40,78 @@ def start(self):
...
             input=json.dumps({"job_id": job_id}),
         )
 
     def process_finished_jobs(self) -> Dict:
         """
-        Scheduled checker to check the in flight jobs, validate if they've finished,
-        and if they have, push them through to LR-10 (step function)
+        Scheduled checker to check the in flight jobs status.
+        If they have finished, push them through to LR-10 (step function)
+        If they have passed the cutoff time, delete the job and raise a log that can be alerted on
```
josh-scargill-nhsd, line 80: Maybe just worth clarifying here, you aren't deleting the job, but the jobs record specifically in the InFlights table.

packages/lambda_code/LR_09_scheduled_check/lr_09_lambda_handler.py
```diff
@@ -40,67 +40,78 @@ def start(self):
...
+                        "logger": "Schedule check",
+                        "level": "CRITICAL",
+                        "message": f"JobID: {item.JobId}, status: TIMED_OUT {JobStatus.TIMED_OUT.value}",
+                    },
+                )
+                self.update_job_status(item.JobId, JobStatus.TIMED_OUT.value)
+                timed_out_jobs.append(item.JobId)
+            else:
                 skipped_jobs.append(item.JobId)
                 continue
```
josh-scargill-nhsd, line 109: This could just be a coding preference, but i'm not a fan of it been all squished together, the previous spread out layout was easier to read imo

packages/utils/statuses.py
```diff
@@ -5,4 +5,5 @@ class JobStatus(Enum):
     ADDED_TO_QUEUE = "1"
     PDS_FHIR_API_PROCESSED = "2"
     DEMOGRAPHICS_DIFFERENCES_PROCESSED = "3"
+    TIMED_OUT = "98"
     CLEANED_UP = "99"
```
josh-scargill-nhsd, line 9: I'm glad you made this change because me and Tom recently finally defined them. And firstly this got missed out as a status, but also the code needs updating to reflect what we've defined. 
I've added this status now, could you reflect [this](https://nhsd-confluence.digital.nhs.uk/display/PCRM/List+Reconciliation+Solution+Walkthrough#ListReconciliationSolutionWalkthrough-ListReconciliationStatusFlowEngine) table in code?

test/unittests/lambdas/LR_09_scheduled_check/test_scheduled_check.py
```diff
@@ -33,10 +29,10 @@ def test_three_inflight_records_one_processed_correctly_two_skipped(
 ):
     # Assert values are as expected before processing jobs
     with pytest.raises(StopIteration):
-        before_job_stats = JobStats.query(JOB_ID)
+        before_job_stats = JobStats.query(JOB_ID[1])
         before_job_stats.next()
 
-    before_job = Jobs.query(JOB_ID)
+    before_job = Jobs.query(JOB_ID[1])
     assert before_job.next().StatusId == JobStatus.ADDED_TO_QUEUE.value
```
josh-scargill-nhsd, line 36: As your change affects the job status for `JOB_ID[4]`, I think there should be a test to assert that status has changed before/after too

test/unittests/packages/utils/test_datetimezone.py
```diff
@@ -82,10 +82,10 @@ def test_localize_date_with_tz_returns_correct_date():
...
     expected_month = 5
     expected_day = 7
-    expected_hour = 19
+    expected_hour = 18
     expected_minute = 30
     expected_second = 0
-    expected_tz = "Europe/London"
+    expected_tz = "UTC"
 
     actual = localize_date(datetime.now(timezone("US/Eastern")))
```
josh-scargill-nhsd, line 90: Can we have this test at least 2, one of which is what we care about? Potentially cater to either side of the BST time changes just to make sure it's behaving as expected

jamesnunnnhs (approved):

PR #116, opened 2021-08-09
PCRM-3662 Manage Sensitive Patients

aparimoo-nhs:
Process Sensitive Patients 
Add "security" field in DSA payload
Create DSA payload with  Demographic difference as empty list for sensitive patients
Exclude Sensitive patients from registration process


jamesnunnnhs (approved):
lgtm

HarveyNHS (approved):
Approved

abbas-khan8 (approved):

packages/lambda_code/LR_12_pds_registration_status/lr_12_lambda_handler.py
```diff
@@ -117,26 +130,26 @@ def get_pds_exclusive_registrations(self, job_id: str) -> Dict[str, str]:
...
-                        "ADD 5": pds_record["address"][4],
-                        "POSTCODE": pds_record["postcode"],
-                        "TITLE": ", ".join(pds_record["title"]),
-                        "SEX": pds_record["gender"],
-                        "DATE ACCEPT.": datetime.strptime(
-                            pds_record["gp_registered_date"], "%Y-%m-%d"
-                        ).date(),
-                    }
-                )
+                if pds_record:
```
abbas-khan8, line 133: Should this be indented as part of the except block?

aparimoo-nhs: No, in except block we re raise exception so this will never get executed.  API can return empty dict for non exist this check ensures not have key errors when we call `row.append`

PR #123, opened 2021-08-11
Adding config for NFT Env

joeds13:


joeds13:
Will recreate when needed

PR #135, opened 2021-08-16
PCRM-999 Demo fixes

aparimoo-nhs:
API retry fixes
LR-07 is comparison set to true when gp match is done
Update condition to filter out unrestricted patients in LR11


packages/utils/pds_api_service.py
```diff
@@ -254,8 +258,8 @@ def parse_json_data(self, field: str) -> str:
             return item.value
 
     def process_status_4xx_responses(
-        self, status_code: int, nhs_number: str, job_id: uuid
-    ) -> Union[Dict[str, str], PDSAPIError]:
+        self, response, status_code: int, nhs_number: str, job_id: uuid
+    ) -> Union[Dict[str, str], PDSAPIError, None]:
```
jamesnunnnhs, line 262: PDSAPIError isn't a return val, just need to document it in a `Raises` section in the docstring

jamesnunnnhs (approved):

joeds13 (changes requested):
Please do not include a previous or future ticket number in either the commit message or branch name

A ticket number is no longer required in the branch name for a build to succeed, if you really want a ticket number use `PCRM-0000`

PR #136, opened 2021-08-16
PCRM-8888 Add pytest artifacts in workflow

aparimoo-nhs:
Generate unit test report.html artifacts


jamesnunnnhs (approved):

joeds13 (changes requested):
Please do not include a future ticket number in either the commit message or branch name

A ticket number is no longer required in the branch name for a build to succeed, if you really want a ticket number use `PCRM-0000`

PR #138, opened 2021-08-16
PCRM-0000 Demo fixes

aparimoo-nhs:
PDS API Retry refactor
LR-12 fix for address where address is not always has 5 items in list
LR-12 fix for Date accept where PDS may not send date
LR-11 fix where filter is for non sensitive patients instead of unrestricted


joeds13 (changes requested):
The commit message still contains PCRM-9999, please remove it or use PCRM-0000

packages/utils/pds_api_service.py
```diff
@@ -125,16 +133,11 @@ def get_access_token(self):
...
-                        f"{self.ssm_store_path}{PDSParamStore.PDS_ACCESS_TOKEN.value}",
-                        region="eu-west-2",
-                        data_string=token,
-                        string_type="SecureString",
-                    ),
-                    retry_on_exception=RuntimeError,
+                put_ssm_params(
+                    f"{self.ssm_store_path}{PDSParamStore.PDS_ACCESS_TOKEN.value}",
+                    region="eu-west-2",
+                    data_string=token,
```
josh-scargill-nhsd, line 139: We originally had a retry on this, since we don't now will it throw uncaught errors?

packages/utils/pds_api_service.py
```diff
@@ -254,65 +258,78 @@ def parse_json_data(self, field: str) -> str:
...
         """
         try:
             token = json.loads(self.ssm_params[PDSParamStore.PDS_ACCESS_TOKEN.value])
-        except JSONDecodeError as e:
+            time_now = datetime.datetime.now()
+            token_issue_time = str(token["issued_at"])[:10]
+            token_expire_in_seconds = int(token["expires_in"])
+            issue_datetime = datetime.datetime.fromtimestamp(int(token_issue_time))
+            token_expiry_time = issue_datetime + datetime.timedelta(0, token_expire_in_seconds)
+            return time_now > token_expiry_time
```
josh-scargill-nhsd, line 329: Assuming this is the best way to do this, I wonder if we should move it to a function to just check if expired?

packages/lambda_code/LR_11_gp_registration_status/lr_11_lambda_handler.py
```diff
@@ -59,7 +60,13 @@ def get_gp_exclusive_registrations(self, job_id: str) -> Success:
...
                 Demographics.GP_RegistrationStatus != GPRegistrationStatus.MATCHED.value
             )
-            & (Demographics.PDS_Sensitive == "U"),
+            & ~(
+                Demographics.PDS_Sensitive.is_in(
+                    SensitiveMarkers.RESTRICTED.value,
+                    SensitiveMarkers.VERY_RESTRICTED.value,
+                    SensitiveMarkers.REDACTED.value,
+                )
+            ),
```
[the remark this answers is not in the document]
aparimoo-nhs: yes but earlier filter was on `U` as we have seen they are sending records with no markers, this ensures we are not processing any sensitive patients but allowing `U``and blank ones

packages/utils/pds_api_service.py
aparimoo-nhs: Since it raises `toomanyupdates` it means other lambdas have already updated so no need to retry

packages/utils/pds_api_service.py
```diff
@@ -254,65 +258,95 @@ def parse_json_data(self, field: str) -> str:
...
+                # Patient Record is REDACTED
+                redacted_patient_data = self.convert_pds_to_list_rec_data()
+                redacted_patient_data.update(
+                    {"sensitive": SensitiveMarkers.REDACTED.value}
+                )  # update dict with REDACTED
+                return redacted_patient_data
+
+            elif status_code == 404 and error_code == "RESOURCE_NOT_FOUND":
+                # PDS data not found for the patient
+                return
```
jamesnunnnhs, line 296: I think this might as well not be defined since there are no more conditions that are specifically handled. You could be explicit and just do 
```
if status_code == 404 and error_code == "INVALIDATED_RESOURCE":
    # Patient Record is REDACTED
    redacted_patient_data = self.convert_pds_to_list_rec_data()
    redacted_patient_data.update(
        {"sensitive": SensitiveMarkers.REDACTED.value}
    )  # update dict with REDACTED
    return redacted_patient_data

else: 
    return
```

or just leave the `else` out

packages/utils/pds_api_service.py
```diff
@@ -254,65 +258,95 @@ def parse_json_data(self, field: str) -> str:
...
+            time_now = datetime.datetime.now()
+            token_issue_time = str(token["issued_at"])[:10]
+            token_expire_in_seconds = int(token["expires_in"])
+            issue_datetime = datetime.datetime.fromtimestamp(int(token_issue_time))
+            token_expiry_time = issue_datetime + datetime.timedelta(0, token_expire_in_seconds)
+            return time_now > token_expiry_time
+
+        except KeyError:
+            # Token issued doesn't have required field
+            return True
```
jamesnunnnhs, line 352: I'd move the code you're not `try`ing outside the try/except. If ever there's a bug in the other function calls that raises a KeyError, it would be horrendous to track down. Something like this

```
time_now = datetime.datetime.now()
try:
    token_issue_time = str(token["issued_at"])[:10]
    token_expire_in_seconds = int(token["expires_in"])

except KeyError:
    # Token issued doesn't have required field
    return True

issue_datetime = datetime.datetime.fromtimestamp(int(token_issue_time))
token_expiry_time = issue_datetime + datetime.timedelta(0, token_expire_in_seconds)
return time_now > token_expiry_time
```

packages/utils/ssm.py
```diff
@@ -25,11 +26,12 @@ def get_ssm_params(ssm_path, region):
...
+def put_ssm_params(ssm_path, data_string, region, string_type="SecureString"):
     try:
         ssm = ssm_client(region)
         ssm.put_parameter(
             Name=ssm_path, Value=json.dumps(data_string), Type=string_type, Overwrite=True
         )
-    except ssm_client.exceptions.TooManyUpdates:
-        raise RuntimeError("SSM TooManyUpdates")
+    except ClientError:
+        # ignore TooManyUpdates Exceptions as other lambda have already updated the token
```
jamesnunnnhs, line 36: Maybe worth clarifying in the comment why you're catching `ClientError` rather than `TooManyUpdates` for this

packages/lambda_code/LR_11_gp_registration_status/lr_11_lambda_handler.py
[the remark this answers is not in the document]
josh-scargill-nhsd: Oh hang on, just spotted the `~` which i'm assuming is a "not this"

jamesnunnnhs (approved):

abbas-khan8 (approved):

packages/utils/pds_api_service.py
```diff
@@ -21,6 +21,14 @@ class PDSAPIError(Exception):
     pass
 
 
+class ExpiredTokenError(Exception):
```
HarveyNHS, line 24: Are we also handling 500 errors aswell?

HarveyNHS (commented):
Please review

packages/utils/pds_api_service.py
aparimoo-nhs: Yes, status code other than 200/4xx will raise PDSAPI Exception

packages/utils/pds_api_service.py
```diff
@@ -102,7 +110,7 @@ def get_access_token(self):
         """
 
         try:
-            if self._is_token_expired():
+            if self._is_token_valid():
```
josh-scargill-nhsd, line 113: To clarify, is this right?
The name would suggest it only enters this code if true.
```
Is token valid? 
    yes? recheck the token
    no? don't need to recheck the token
```

packages/utils/ssm.py
```diff
@@ -25,11 +26,14 @@ def get_ssm_params(ssm_path, region):
...
-def put_ssm_params(ssm_path, data_string, string_type, region):
+def put_ssm_params(ssm_path, data_string, region, string_type="SecureString"):
     try:
         ssm = ssm_client(region)
         ssm.put_parameter(
             Name=ssm_path, Value=json.dumps(data_string), Type=string_type, Overwrite=True
         )
-    except ssm_client.exceptions.TooManyUpdates:
-        raise RuntimeError("SSM TooManyUpdates")
+    except ClientError:
```
jamesnunnnhs, line 35: @joeds13

joeds13 (approved):

PR #141, opened 2021-08-16
Python 3.9

joeds13:
Make use of the Python 3.9 Lambda Runtime


josh-scargill-nhsd:
Don't forget to update documentation when this goes in

PR #154, opened 2021-08-19
Load test gp flat file - volume test 100k

jamesnunnnhs:


terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -16,7 +16,7 @@ locals {
   }
 
   pds_fhir_api_url = {
-    default = "https://sandbox.api.service.nhs.uk",
+    default = "https://ref.api.service.nhs.uk",
```
jamesnunnnhs, line 19: Change this back to sandbox

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -28,6 +28,7 @@ def initialise(self):
         pass
 
     def start(self):
+        print(self.event.raw_event)
```
jamesnunnnhs, line 31: @joeds13 this is printing the raw event dict, it contains the event type doc'ed here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html#supported-notification-event-types

packages/lambda_code/LR_07_pds_hydrate/lr_07_lambda_handler.py
```diff
@@ -45,6 +46,14 @@ def start(self):
...
         self.job_id = str(body["job_id"])
         self.log_object.set_internal_id(self.job_id)
+        self.log_object.write_log(
+            "LR07I05",
+            log_row_dict={
+                "patient_id": body["id"],
+                "nhs_number": body["NHS_NUMBER"],
+                "job_id": self.job_id,
+            },
+        )
```
jamesnunnnhs, line 56: This is to log the lambda invocation as soon as possible with a patient id so we can see if it's triggered from s3

PR #192, opened 2021-09-03
PCRM-4311: Normalisation of data during comparisons

johnwaghorn:


test/unittests/packages/listrec_comparison_engine/test_comparison_formats.py
```diff
@@ -0,0 +1,98 @@
+from listrec_comparison_engine import utils
+
+
+def test_format_gp_dob_happy():
```
joeds13, line 4: It would be good to utilise `pytest`s `parametrized` tests, to cover multiple cases easily, like used here: 
https://github.com/NHSDigital/list-reconciliation/blob/develop/test/unit/gp_file_parser/test_validators.py

packages/listrec_comparison_engine/utils.py
```diff
@@ -1,3 +1,6 @@
+from datetime import datetime
+
+
 def strip_ws(in_str: str) -> str:
```
joeds13, line 4: `in_str: str` is a bit of a stutter, if we have the type hinting the variable could just be `input`

packages/listrec_comparison_engine/utils.py
```diff
@@ -1,12 +1,84 @@
-def strip_ws(in_str: str) -> str:
+from datetime import datetime
```
joeds13, line 1: This file has been superseded by `packages/listrec_comparison_engine/format.py` so should be removed from the PR

joeds13 (changes requested):

joeds13 (approved):

joeds13 (dismissed):

joeds13 (approved):

PR #209, opened 2021-09-14
PCRM-3971 Test Data Generator

joeds13:
Generate PDS, DPS PDS and GP data


joeds13:
Would like to get this merged as is, it's not in use anywhere yet, so the in progress-ness of it will have no affect.

Thinking about how we will piece the GP File and DPS PDS data together with the PDS API responses, and doing that at a representative scale (~10,000 record files), we appear to need to have a mocked PDS API that we can back with the same dataset and derive all the test files from it. 

So a new PR with the mocked API and this generator used to create the backing data for it is the next steps.

johnwaghorn (approved):

PR #227, opened 2021-09-20
PCRM-4253-getting-ssm-tokens

johnwaghorn:


johnwaghorn:
Closing PR following a meeting with Alex Hawdon, system will be redesigned to avoid using FHIR API.

johnwaghorn:
Closing PR following a meeting with Alex Hawdon, system will be redesigned to avoid using FHIR API.

PR #229, opened 2021-09-21
Decouple Lambdas in Terraform

joeds13:
Constructing the ARNs, rather than passing them from the target Lambdas
module, means the terraform modules do not depend on each other and can
be deployed in parrallel, this will improve the deployment speed and
gain back some of the time lost from moving to VPC deployed Lambdas


johnwaghorn (dismissed):

johnwaghorn (dismissed):

joeds13:
Closing as we couldn't easily create an improvement, a single Lambda has been seen taking >29 minutes to deploy, with an average build time of 35 minutes, nothing we can do can make more that a five minute improvement

PR #237, opened 2021-09-23
PCRM-4312-rewrite-email-sending-to-use-separate-lambda

johnwaghorn:


packages/lambda_send_email/lambda_handler.py
```diff
@@ -0,0 +1,119 @@
...
+                log_row_dict={
+                    "logger": "SendEmail",
+                    "level": "INFO",
+                    "message": f"Got object {key} from {self.bucket}",
+                },
+            )
+            return response
+        except Exception as e:
+            self.log_object.write_log(
+                "LR_send_email",
```
joeds13, line 61: Log IDs need to be paired with a log in cloud log config

packages/lambda_send_email/lambda_handler.py
```diff
@@ -0,0 +1,119 @@
...
+                log_row_dict={
+                    "logger": "SendEmail",
+                    "level": "CRITICAL",
+                    "message": f"Error getting object {key} from bucket {self.bucket}. Make sure they exist and your "
```
joeds13, line 65: Use key/values for error information, as per cloud log config

packages/send_email/send.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class GlobalArgs:
+    pass
+
+
+global_args = GlobalArgs()
+
+
+def send(service_index: str, to: list, subject: str, body) -> bool:
```
joeds13, line 13: As this is the main package interface, expose it through the `__init__.py` for easy consumption

Also add a ReadMe, supporting adoption and detailing requirements. 

What is a service index, does that name make sense if used by a non-ListRec project?

Is it worth passing in the `EMAIL_BUCKET` to this interface as opposed to it being a non-explicit requirement of calling this function? Can you try some sort of bucket lookup, if no `EMAIL_BUCKET` is given?

packages/send_email/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        return True
+
+    else:
+
+        EMAIL_BUCKET = os.environ["EMAIL_BUCKET"]
+
+        s3 = boto3.client("s3")
+
+        try:
+            retry_func(
```
joeds13, line 39: Is this required? I believe the AWS SDK has some level of retries built in?

Can we just return an error and let the calling function decide to retry or not?

packages/send_email/utils.py
```diff
@@ -0,0 +1,81 @@
...
+                lambda: s3.put_object(Body=email, Bucket=EMAIL_BUCKET, Key=filename),
+                wait_exponential_multiplier=1000,
+                wait_exponential_max=10000,
+                stop_max_attempt_number=10,
+            )
+
+        except ClientError as e:
+            logger.write_log(
```
joeds13, line 47: Packages should not log, logging is business logic and is a higher level.

We should just raise Errors/Exceptions here that the calling code can choose to handle, reducing coupling

packages/send_email/utils.py
```diff
@@ -0,0 +1,81 @@
...
+        }
+    )
+
+
+def generate_filename(service_index, subject) -> str:
+    prefix = service_index.upper()
+    timestamp = datetime.now().strftime("%Y%m%d%H%M%S")
+    suffix = " ".join(subject.split()[:3]).lower()
+    suffix = suffix.replace(" ", "_")
+    filename = str(prefix + "_" + timestamp + "_" + suffix + ".json")
```
joeds13, line 78: Use format strings not string concatenation

packages/send_email/utils.py
```diff
@@ -0,0 +1,81 @@
...
+            "service_index": service_index.upper(),
+            "timestamp": timestamp,
+            "to": to,
+            "subject": subject,
+            "body": body,
+        }
+    )
+
+
+def generate_filename(service_index, subject) -> str:
```
joeds13, line 73: Do we need all this information in the file name?

What happens if we want a date in the subject, in the usual UK format of day/month/year? We should normalise/remove invalid characters oppose to rejecting the entire email due to constraints that are not constraints of sending a normal email.

However, as all the relevant info is in the email, we could just use a GUID as the filename and not need any of this?

packages/send_email/utils.py
```diff
@@ -0,0 +1,81 @@
...
+                    "level": "Critical",
+                    "message": f"Failed to upload {filename} to {EMAIL_BUCKET} due to {e}",
+                },
+            )
+
+            return False
+        return True
+
+
+def format(service_index, to, subject, body):
```
joeds13, line 60: The name `format` doesn't really tell me what this is doing, generally most function names could look to be revised to be a bit more descriptive

Also, no type hinting for args or returns in this file

packages/send_email/utils.py
```diff
@@ -0,0 +1,81 @@
+import json
```
joeds13, line 1: `utils.py` is not a great name, its best to avoid them and make the naming more explicit

https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

packages/lambda_send_email/lambda_handler.py
```diff
@@ -0,0 +1,119 @@
...
+            "email_addresses": json_content["to"],
+            "subject": json_content["subject"],
+            "message": json_content["body"],
+        }
+
+    def send(self, key):
+
+        email = self.generate_email(key)
+
+        nhs_mail_relay.send_email(
```
joeds13, line 91: Probably worth moving the couple of functions out of `nhs_mail_relay` into `send_email`, just to keep it all nicely packaged and with the idea in mind that it could be shared with other NHSD/Spine AWS projects

joeds13 (changes requested):

.gitignore
```diff
@@ -17,6 +17,8 @@ build/*
 
 # Test Data
 test/_data/generator/tmp/*
+test/integrationtests/reports/*
```
joeds13, line 20: These aren't required as we have removed Gauge from the project

packages/send_email/ReadMe.md
```diff
@@ -0,0 +1,43 @@
...
+pip install boto3, botocore, exchangelib
+```
+
+## Usage
+
+```python
+from send_email.send import to_outbox
+
+service = "spine service"
+to = ["bob@axes.com"]
```
joeds13, line 23: Please use obviously test email addresses, throughout the documentation and tests

Something like: `example_email@nhs.net`

packages/send_email/ReadMe.md
```diff
@@ -0,0 +1,43 @@
...
+body = "a long body"
+
+to_outbox(service, to, subject, body)
+# returns 'boolean'
+```
+
+```python
+from send_email.send import send
+
+username = "admin"
```
joeds13, line 34: Again, obviously test values, just to be safe

Like: `example_username` and `example_password`

packages/send_email/send.py
```diff
@@ -0,0 +1,44 @@
...
+            * User Principal Name (UPN)
+        password: Clear-text password`
+        event: {
+            "email_addresses": ["rand@nhs.com", "rands@nhs.com"],
+            "subject": subject,
+            "message": body,
+            }
+
+    """
+    send_email(username, password, event)
```
joeds13, line 44: We need some code here to handle the potential exceptions raised by `send_email`

Bu not sure if actually makes more sense to lift up `send_email` to here, as all we're currently doing is directly "proxying" all the args to that function, which seems a bit superfluous

joeds13 (changes requested):
Are you planning on doing the deployment/terraform code for the sending Lambda in a follow up PR?

packages/lr_logging/list_reconciliation_logs.cfg
```diff
@@ -285,3 +285,23 @@ Log Text = Failed to cleanup record in table='{inflight_table}' for jobId='{job_
...
+Log Level = INFO
+Log Text = Deleted object='{key}' from bucket='{bucket}'
+
+[LRSEC01]
+Log Level = CRITICAL
+Log Text = Error getting object='{key}' from bucket='{bucket}'. Make sure they exist and your bucket is in the same region as this function
+
+[LRSEC02]
+Log Level = CRITICAL
+Log Text = Error deleting object='{key}' from bucket='{bucket}'.
```
joeds13, line 307: Wonder why this missing file ending newline hasn't been picked up by `pre-commit`? 🤔

.gitignore
```diff
@@ -17,6 +17,7 @@ build/*
 
 # Test Data
 test/_data/generator/tmp/*
+test/integrationtests/.gauge/*
```
joeds13, line 20: This one too, as it's also from Gauge

joeds13 (changes requested):

joeds13 (approved):

PR #276, opened 2021-10-19
pcrm-4523-LR-04-to-send-emails-via-the-send_email-package-and-lambda

johnwaghorn:
…ctor, terraform changes


packages/lr_04_feedback_failure/lr_04_lambda_handler.py
```diff
@@ -185,29 +209,26 @@ def validate_log(self):
...
+            self.sent = to_outbox(service, [to], subject, body, bucket)
+            self.log_object.write_log(
+                "LR04I02",
+                log_row_dict={
+                    "email_address": self.to,
+                    "upload_filename": self.log["file"],
+                    "job_id": self.job_id,
+                },
+            )
+        except:
```
antonyibbotson1-nhs, line 228: Isn't this going to swallow the exception? And we'll lose the stack trace? The custom exception below is fine for telling us that it failed, but I think we'll lose **why** it failed

antonyibbotson1-nhs (approved):

adrianwalker5-nhs (approved):

robertnixson1-nhs (approved):

PR #317, opened 2021-11-22
PCRM-4582-send-to-dps

johnwaghorn:


packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def __init__(self):
+        super().__init__(additional_log_config=get_cloudlogbase_config())
+        self.job_id = None
+        self.s3 = boto3.client("s3")
+        self.mesh_params = get_ssm_params(
+            self.system_config["MESH_SSM_PREFIX"], self.system_config["AWS_REGION"]
+        )
+
+    def initialise(self):
```
antonyibbotson1-nhs, line 32: empty method?

terraform/stacks/list-reconciliation/s3.tf
```diff
@@ -12,25 +12,46 @@ module "lr_01_gp_extract_input" {
 
     {
       events              = ["s3:ObjectCreated:*"]
-      lambda_function_arn = module.lr_04_feedback_failure.lambda.arn
-      key_prefix          = "fail/"
+      lambda_function_arn = module.lambda_send_to_dps.lambda.arn
+      key_prefix          = null
```
antonyibbotson1-nhs, line 16: Will this be null? Might need to do some digging on this one - potentially out-of-scope

lambdas/send_to_dps/main.py
```diff
@@ -0,0 +1,7 @@
+from lambda_send_to_dps.lambda_handler import SendToDps
```
antonyibbotson1-nhs, line 1: I'm assuming contextually we've agreed not to rename this given it was already in-flight when the other renaming tickets were raised?

packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -0,0 +1,377 @@
...
+        self.response = success(
+            message="lr_send_to_dps file sent",
+            internal_id=self.log_object.internal_id,
+            job=self.job_id,
+            file=self.input_file,
+        )
+
+    def create_init_job(self):
+
+        """Creates an intial Job item in DynamoDb."""
```
antonyibbotson1-nhs, line 126: typo in docstring

packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -0,0 +1,377 @@
...
+        try:
+            job_item = Jobs(
+                self.job_id,
+                PracticeCode="N/A",
```
antonyibbotson1-nhs, line 130: I'm a bit unsure about this, I don't think "N/A" is suitable in this context, I think I'd rather leave it blank because the practice code is still applicable but we've just been unable to parse the file for x reasons

packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -0,0 +1,377 @@
...
+        )
+
+    def initialise(self):
+        pass
+
+    def start(self):
+        try:
+            self.job_id = str(uuid4())
+            self.input_bucket = self.event["Records"][0]["s3"]["bucket"]["name"]
+            self.input_file = self.event["Records"][0]["s3"]["object"]["key"]
```
antonyibbotson1-nhs, line 43: What happens if we have two objects added at the same time, are the records munged into a single event or does it trigger two separate ones?

packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -0,0 +1,377 @@
...
+                j.save()
+            self.log_object.write_log(
+                "LRSDI03",
+                log_row_dict={
+                    "upload_filename": self.input_file,
+                    "job_id": self.job_id,
+                },
+            )
+        except (PynamoDBConnectionError, PutError) as e:
+            self.handle_extract(InputFolderType.RETRY)
```
antonyibbotson1-nhs, line 238: Where is self.handle_extract coming from?

packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -0,0 +1,377 @@
...
+            self.log_object.write_log(
+                "LRSDI05",
+                log_row_dict={
+                    "upload_filename": self.input_file,
+                    "job_id": self.job_id,
+                    "status_code": status_code,
+                },
+            )
+        except (PynamoDBConnectionError, PutError) as e:
+            self.handle_extract(InputFolderType.RETRY)
```
antonyibbotson1-nhs, line 274: Where is self.handle_extract coming from?

packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -0,0 +1,377 @@
...
+        self.update_job_info(practice_code, num_of_records)
+
+        # send valid file via mesh
+        self.send_mesh_file(self.filename, file)
+
+        # once the file has been sent to DPS, update job item in db
+        self.update_job_status("SENT_TO_DPS")
+
+        # cleanup input files
+        self.cleanup_files(self.input_bucket, self.input_file)
```
antonyibbotson1-nhs, line 114: This probably wants to be removed? https://nhsd-jira.digital.nhs.uk/browse/PCRM-4177

antonyibbotson1-nhs: We could handle this case (separate ticket) using S3 lifecycle rules anyway.

packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -0,0 +1,377 @@
...
+
+        # respond with error
+        self.response = success(
```
antonyibbotson1-nhs, line 117: This comment is a bit misleading, is this not the success response?

packages/lambda_send_to_dps/lambda_handler.py
```diff
@@ -19,8 +19,8 @@
 class SendToDps(LambdaApplication):
     """
     Lambda:
-        responsible for responsible for creating a Job, tracking it DynamoDB and then sending a given input file off to DPS
-        if the input file is invalid, then reject
+        Responsible for responsible for creating a Job, tracking it DynamoDB and then sending a given input file off to DPS.
```
antonyibbotson1-nhs, line 22: Responsible for responsible for responsible for responsible for responsible for responsible for

antonyibbotson1-nhs (approved):

PR #328, opened 2021-12-08
PCRM-4835 - update default tf workspace on main and add mesh module r…

antonyibbotson1-nhs:
…eadme


terraform/stacks/mesh/README.md
```diff
@@ -1,3 +1,11 @@
 # Mesh Stack
 
-Mesh send/recieve Lambdas
+## Mesh send/receive Lambdas
+
+The majority of the terraform is stored in another project within the spine-core-aws-common project which can be
+found [here](https://github.com/NHSDigital/spine-core-aws-common).
+
+Performing a [`terraform get`](https://www.terraform.io/docs/cli/commands/get.html) should pull the project files into this space within the `.terraform` directory making it
```
Hobdell, line 8: Is this something that should be done when first opening up the repo in your IDE?

Hobdell: It says it _should_  pull the project files in, does this command often fail?

antonyibbotson1-nhs: 1) You don't have to, but you can - it'll make life a bit easier as you won't need to switch between projects to look at the code.
2) The only time it should fail is if you do not have access to that repo or other github/terraform issues (rare).

Hobdell: 👍🏻 In that case,like you suggested, I'd add a comment in the onboarding docs that it'd be useful when working on that module.

johnwaghorn (approved):

PR #332, opened 2021-12-13
PCRM-4866 - add separate lists for subnet ids in local and target bas…

antonyibbotson1-nhs:
…ed on environment in list-rec datafile


terraform/stacks/list-reconciliation/locals.tf
```diff
@@ -15,6 +15,24 @@ locals {
     # prod = "arn:aws:iam::000000000000:role/LRTerraformDeploy"
   }
 
+  intra_subnets = {
+    "dev" : ["dev-vpc-intra-eu-west-2a", "dev-vpc-intra-eu-west-2b", "dev-vpc-intra-eu-west-2c"]
```
Hobdell, line 19: Why hard-code dev when preprod and prod use a local variable?

Hobdell: local.environment is the terraform workspace that needs to use the shared resources in the dev account.

Hobdell (approved):

PR #354, opened 2022-01-12
Pcrm 4853 tagging in behave

Hobdell:


.github/workflows/create-environment.yml
```diff
@@ -215,3 +222,10 @@ jobs:
           name: integration-test-report
           path: ./build/behave/*
           retention-days: 7
+      - name: Upload Allure test result artifacts
```
johnwaghorn, line 225: duplicate of line 55

Hobdell: The reason behind the duplication: if create-python-package and  run-integration-tests stages were triggered separately, the command would need to exist in both

johnwaghorn (approved):
discussed and reviewed locally - looks good

PR #358, opened 2022-01-13
PCRM-4856 Ensure validate_and_forward lambda can handle both UTF-8 & latin encoding

LukeC92:


test/_data/unit/A82096_GPR4LNA1.CSA
```diff
@@ -0,0 +1,5 @@
+503\*
+DOW~1~1111111,1234~LNA~20210328~1340~1557492~1234567890~SOMEBODY~JOHN~SOMEONE~MR~1~20020101~FL?T A~THE STREET
```
johnwaghorn, line 2: can you double check the encoding for two of these files? I'm reading charset=us-ascii instead of charset=iso-8859-1

for: A82096_GPR4LNA1.CSA & A82097_GPR4LNA1.CSA

The other files have the correct encoding.

johnwaghorn (dismissed):
Looks good. Tfsec issue will need to be looked at to get this merged in.

johnwaghorn (approved):

PR #363, opened 2022-01-17
PCRM-4812 spike eventbridge

johnwaghorn:


johnwaghorn:
EventBridge module created here: https://github.com/NHSDigital/list-reconciliation/pull/371

PR #369, opened 2022-01-20
PCRM-5058 - add ingress rule from private subnet to vpc endpoints for…

antonyibbotson1-nhs:
… aws services


terraform/stacks/list-reconciliation/lambdas.tf
```diff
@@ -188,7 +188,6 @@ module "validate_and_forward" {
     JOBS_TABLE          = module.lr_03_jobs.table.name
     MESH_SSM_PREFIX     = "/${local.environment}/mesh/"
     LR_13_OUTPUT_BUCKET = module.lr_13_output.bucket.bucket
-
-
+    MESH_BUCKET         = "${local.environment}-mesh-bucket"
```
johnwaghorn, line 191: could this be module.mesh_bucket.bucket.bucket just incase the bucket name changes

Hobdell (dismissed):

terraform/stacks/list-reconciliation/lambdas.tf
antonyibbotson1-nhs: Sure

johnwaghorn (approved):

PR #374, opened 2022-01-24
Pcrm 5006 retospective tagging

Hobdell:


LukeC92:
This all looks good to me, the one thing I'm wondering is why a the 'self' argument has been added to all the tests?

LukeC92 (approved):

PR #379, opened 2022-01-26
PCRM-5114-eventbridge-send-email

johnwaghorn:


terraform/stacks/list-reconciliation/eventbridge.tf
```diff
@@ -0,0 +1,55 @@
+module "eventbridge_send_email" {
```
antonyibbotson1-nhs, line 1: Name probably doesn't need to be prefixed with "eventbridge_" as it's in the eventbridge.tf file .
I'd be tempted to explicitly state it's the send_email lambda though, so:
`send_email_lambda`
Just because it could be quite ambiguous as there is SES in AWS.
Adding "lambda" as a suffix will also make it clearer when looking at rules in the console just in-case we have further use-cases that aren't used with lambda or S3 etc.

antonyibbotson1-nhs (approved):

PR #382, opened 2022-01-26
PCRM-5140 cloudtrail-bucket-dynamic-force-destroy

johnwaghorn:


johnwaghorn:
no longer necessary as we handling the cloudtrail and it's bucket differently now: https://github.com/NHSDigital/list-reconciliation/pull/385

PR #385, opened 2022-01-27
PCRM-5140 cloudtrail module moved to the account stack and send-email lambda eventbridge trigger

johnwaghorn:


terraform/modules/cloudtrail_s3_logs/cloudtrail_s3logs_bucket.tf
```diff
@@ -1,5 +1,5 @@
 resource "aws_s3_bucket" "cloudtrail_s3logs" {
-  bucket = "${var.environment}-${local.name}-ct-logs"
+  bucket = "${var.environment}-ct-logs"
```
Hobdell, line 2: Is there a char limit when naming resources? Could the name be made more explicit e.g. "${var.environment}-cloudtrail-logs"

Hobdell (approved):

PR #393, opened 2022-02-01
PCRM-5179 dps-json-log-specification

johnwaghorn:


docs/specifications/dps-json-log-specification.md
```diff
@@ -0,0 +1,34 @@
+# DPS JSON Log Specification
+
+Once DPS has completed a comparison, three csv files defined here [output](output-specification.md), will be sent to the dps_inbound bucket. A json log file should accompany the three csv files.
```
Hobdell, line 3: Can a line be added to state this log file contains metadata about the three files and should always be written last so the number of records in each file is correct at time of writing.

Hobdell (approved):

PR #399, opened 2022-02-03
PCRM-5182 dps diagrams

johnwaghorn:


Hobdell (commented):
The  PDS Delta table should be renamed as pds.full, this is the copy of the PDS extract from Spine that is updated daily via S3. There is another PDS extract that is delivered to DPS via MESH that fills a table called pds.pds which in theory is updated daily but this is being questioned if it's actually happening.

Hobdell (approved):

PR #414, opened 2022-02-10
PCRM-5276 - modify makefile cmds and add run book for deploying to pr…

antonyibbotson1-nhs:
…eprod


Makefile
```diff
@@ -165,3 +166,24 @@ get-branch-env:
...
+checkout-version:
+	git fetch --tags
+	git checkout $(version)
+
+
+#
+# aws-vault
+#
+
+aws-vault-dev:
```
Hobdell, line 179: nice to have: function to clear aws vault

docs/running-preprod-terraform-locally.md
```diff
@@ -0,0 +1,81 @@
+# Deployment to preprod
```
Hobdell, line 1: A directory should be created for numbered KOPs under docs. In Confluence there are two docs, one for deploying to Pre-prod and manual deployment to dev / workspace - both should exist in the repo. 

Question - should we be mirroring the docs or prioritise PCRM-5096 to prevent work duplication...?

.github/workflows/destroy-environment.yml
```diff
@@ -24,14 +24,14 @@ jobs:
...
       - name: Terraform Init
         run: |
-          make init
-          make workspace env=$(make get-branch-env branch=${{ github.event.ref }})
+          make terraform-init
+          make terraform-workspace workspace=$(make get-branch-env branch=${{ github.event.ref }})
       - name: Terraform Destroy
         run: |
-          make destroy
+          make terraform-destroy
```
Hobdell, line 31: more descriptive names ❤️

Makefile
antonyibbotson1-nhs: Tried that, couldn't get it to work properly.

docs/running-preprod-terraform-locally.md
antonyibbotson1-nhs: I had considered adding that structure, but this feels more of a guide rather than a KOP. These docs will be changed in the near future regardless.

I don't necessarily feel that "deploying to dev" is worthy of adding to the repo, given that the expected workflow is to use the pipeline to deploy.

Pushing the docs to confluence is definitely worth it, however more for static documentation rather than a deployment KOP which is likely to change.

Makefile
antonyibbotson1-nhs: Scratch that, got it working

Hobdell (approved):

PR #415, opened 2022-02-10
PCRM-4813 review-parsing-and-validations-unit-test-cases

johnwaghorn:


Pipfile
```diff
@@ -52,7 +52,7 @@ python_version = "3.9"
 
 [scripts]
 # Testing
-unittests = "pytest -v --doctest-modules --cov=packages --html=build/pytest/unit_tests_report.html --alluredir=build/allure/"
+unittests = "pytest -vv --doctest-modules --cov=packages --html=build/pytest/unit_tests_report.html --alluredir=build/allure/"
```
Hobdell, line 55: is the verbosity level inc. on purpose or a hangover from debugging?

johnwaghorn: Yes left on purpose as it will only come into effect when there is an issue with a test so it's nice to have.

test/unit/gp_file_parser/test_validators.py
```diff
@@ -430,7 +438,18 @@ def test_validate_record_contains_invalid_dict(
             ("A", [], ("A", v.INVALID_TRANS_ID)),
             ("1", [2, 3, 4], (1, None)),
             ("1", [1, 2, 3, 4], (1, v.INVALID_TRANS_ID)),
+            (
+                "200000000000000000000000000000000000.000",
```
antonyibbotson1-nhs, line 442: all the 0's

packages/gp_file_parser/file_name_parser.py
```diff
@@ -83,7 +83,6 @@ def process_date(process_datetime: datetime, date_indicator: str) -> datetime:
     """Processes a date from the file name.
         - Date must not exceed the current date
```
Hobdell, line 84: on line 48 the check for the valid_download_name  - make explicit what GPR4 is in the spec/function pls

test/unit/gp_file_parser/test_filename_parser.py
```diff
@@ -328,3 +328,52 @@ def test_validate_filename_invalid_garbled_dat_raises_InvalidFilename(self):
...
+        (
+            ("LV", "20191231"),
+            ("LU", "20191230"),
+            ("LI", "20191218"),
+        ),
+    )
+    def test_process_date_in_new_year_period_for_dec_files_returns_year_minus_one(
+        self, date_indicator, expected
+    ):
+        # LV = December 31 | LI = December 18
```
antonyibbotson1-nhs, line 360: Probably don't need this comment

Hobdell (approved):

PR #418, opened 2022-02-15
PCRM-5308 - update eventbridge instances event_source_arn and amend d…

antonyibbotson1-nhs:
…eployment kops


johnwaghorn (approved):
Discussed offline, looks good to me :)

PR #435, opened 2022-03-03
PCRM-5365 integration test bug

johnwaghorn:


packages/validate_and_forward/lambda_handler.py
```diff
@@ -286,7 +281,6 @@ def send_file_to_dps_outbound(self, file: str):
             )
             raise e
 
-    @retry(exceptions=DoesNotExist, tries=5, delay=10, backoff=2, logger=__name__)
     def update_job_info(self, practice_code: str, num_of_records: int):
```
antonyibbotson1-nhs, line 284: Is this a certainty that the record exists before it tries to update? If DynamoDB chokes or is exhibiting slow responses for ` x `  reason?

packages/gp_file_parser/parser.py
```diff
@@ -283,6 +284,8 @@ def parse_gp_extract_text(
 
         line_number += 2
 
+    if invalid_lines:
+        logger.warning(f"INVALID RECORD(S) FOUND ON LINE PAIRS:::: {invalid_lines}")
```
antonyibbotson1-nhs, line 288: More a general question than anything - Should these log message conform to the standard we're using in the Lambda, with a log reference etc?

packages/validate_and_forward/lambda_handler.py
johnwaghorn: yes and tested with multiple file sizes. It was previously the integration test deleting the record before it could read the value.

antonyibbotson1-nhs: Gotcha

packages/gp_file_parser/parser.py
Hobdell: imo that's overkill for testing

antonyibbotson1-nhs (approved):

PR #446, opened 2022-03-16
PCRM-5121 - Audit and pin python dependencies

LukeC92:


Makefile
```diff
@@ -18,10 +18,15 @@ test: integrationtests unittests
 python-deps:
 	pipenv install --dev --ignore-pipfile
 
+.PHONY:
```
Hobdell, line 21: Phoney targets are usually declared at the start of the file and explicitly state the target, rather than being placed above the target directly. e.g. 

.PHONY: check-safety
check-safety:
        blah

README.md
```diff
@@ -48,6 +48,8 @@ The List Reconciliation Solution is intended to compare the patient list includi
 - [Install Terraform and AWS Prerequisites](./docs/developer_guide/terraform-aws-prerequisites.md)
 - [Deploying to AWS](docs/developer_guide/deploying-to-aws-workspace.md)
 - [Running Integration Tests](./test/integration/README.md)
+- [Running Unit Tests](./test/unit/README.md)
```
Hobdell, line 51: comment on the unit test README, which needs updating and currently isn't part of this changeset:
The tagging with Allure section is incorrect. Unit tests don't have Features and Scenario sections. Epic tags are added at the Class level and ticket tags are added per test.  Can this and the examples be updated.

PR #449, opened 2022-03-17
PCRM-5185 dps-inbound-bucket-and-dps-write-role

johnwaghorn:


terraform/modules/account_policies/dps_write_bucket_role/iam.tf
```diff
@@ -0,0 +1,61 @@
...
+#tfsec:ignore:aws-iam-no-policy-wildcards
+data "aws_iam_policy_document" "DPSWriteToBucketRolePolicyDocument" {
+
+  statement {
+    sid    = "WriteAccessToS3DPSInbound"
+    effect = "Allow"
+    actions = [
+      "s3:PutObject",
+      "s3:ListBucket",
+      "s3:GetObject",
```
Hobdell, line 28: Is this required as DPS should only have the ability to write to this bucket

terraform/modules/account_policies/dps_write_bucket_role/variables.tf
```diff
@@ -0,0 +1,4 @@
+variable "s3_kms_arn" {
+  description = ""
```
Hobdell, line 2: blank description 👀

Hobdell (approved):

PR #458, opened 2022-03-30
PCRM-5716 - install pip on self-hosted runner and add TagSession to a…

antonyibbotson1-nhs:
…ssume role policy


terraform/modules/bootstrap/github_runner/cloudinit.tf
```diff
@@ -280,6 +280,11 @@ write_files:
       unzip terraform_1.0.5_linux_amd64.zip
       sudo mv terraform /usr/local/bin
 
+      echo "<<<<<============+ install python pip +================>>>>"
+      curl -O https://bootstrap.pypa.io/get-pip.py
+      python3.8 get-pip.py --user
```
antonyibbotson1-nhs, line 285: Just need to verify if the cloudinit is ran as root or the GitHub user

